Issue #128, opened 2015-02-04
Unused keys and controllers in modules


atipugin:
There is a problem with keys in controllers that nested in modules. For example:

```shell
$ be i18n-tasks health
Forest (ru) has 99 keys in total. On average, values are 27 characters long, keys have 3.4 segments.
✓ Perfect! No translations are missing.
Unused keys (3) | i18n-tasks v0.7.11
+--------+-----------------------------------+-----------------------------+
| Locale | Key                               | Value                       |
+--------+-----------------------------------+-----------------------------+
|   ru   | projects.campaigns.create.notice  | Кампания успешно создана.   |
|   ru   | projects.campaigns.destroy.notice | Кампания успешно удалена.   |
|   ru   | projects.campaigns.update.notice  | Кампания успешно сохранена. |
+--------+-----------------------------------+-----------------------------+
```

`relative_roots` option is set to:
```yaml
relative_roots:
    - app/views
    - app/controllers
```

Controllers look like this:
```ruby
module Projects
  class CampaignsController < BaseController
    # ...

    def create
      @campaign = @project.campaigns.new(campaign_params)
      authorize(@campaign)
      if @campaign.save
        redirect_to project_campaign_path(@project, @campaign),
                    notice: t('.notice')
      else
        render :new
      end
    end

    # ...
end
```

atipugin:
Added pull request to solve this issue: https://github.com/glebm/i18n-tasks/pull/129

PR #129, opened 2015-02-04
Handle relative keys in controllers nested in modules

atipugin:
Solves this issue: https://github.com/glebm/i18n-tasks/issues/128


glebm:
Thanks, merged!

PR #131, opened 2015-02-17
Add support for the 'it' gem

jwsf:
See https://github.com/iGEL/it for details of the gem.

This change simply tweaks the parser slightly to pick up things like:

    <%= it :your_privacy_is_important, privacy_link: It.link(privacy_url()) %>

...so that it doesn't report erroneous unused translations.

Includes updated specs.

### DIFF
--- a/lib/i18n/tasks/scanners/pattern_scanner.rb
+++ b/lib/i18n/tasks/scanners/pattern_scanner.rb
@@ -70,7 +70,7 @@ module I18n::Tasks::Scanners
     end
 
     def translate_call_re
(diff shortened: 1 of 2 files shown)


glebm:
[it][it] is a great gem, but it is not widely adopted yet.
Support can be achieved with putting this in the config:

```yaml
<% # Adjust the pattern for the it gem support: https://github.com/iGEL/it
I18n::Tasks::Scanners.define_method(:translate_call_re) { 
  /(?<=^|[^\w'\-])i?t(?:ranslate)?/ } %>
```

[it]: https://github.com/iGEL/it

PR #132, opened 2015-03-02
Handle relative keys in controllers whose name consist of multiple words

yujinakayama:
Currently relative keys in `app/controllers/admin_users_controller.rb` are resolved as `admin.KEY`.

With this fix they'll be resolved as `admin_users.KEY` correctly.


glebm:
Thanks! :+1:

PR #135, opened 2015-03-24
Don't require spec_helper where you don't need to.

dleve123:
I was updating my copy of the `I81n` spec and realized
that `require`-ing `spec_helper` wasn't needed to run the spec in isolation.
While loading the `spec_helper` doesn't add much overhead to this project,
requiring `spec_helper` adds a ton of load time for many bloated projects.

Hope this helps!

### DIFF
--- a/spec/i18n_spec.rb
+++ b/spec/i18n_spec.rb
@@ -1,4 +1,3 @@
-require 'spec_helper'
 require 'i18n/tasks'
 
 describe 'I18n' do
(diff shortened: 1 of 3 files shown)


dleve123:
cc/ @glebm 

Let me know if there is anything outstanding items/fixups needed for this PR. Excited to help out a project that has been so useful!

glebm:
Good catch, thank you!

PR #147, opened 2015-05-06
Added option to keep the keys order when using remove-unused

bulinutza:


glebm:
Looks good to me, just needs a test for the new option

lib/i18n/tasks/command/commands/usages.rb
```diff
@@ -31,13 +36,16 @@ def unused(opt = {})
...
             desc: t('i18n_tasks.cmd.desc.remove_unused'),
-            args: [:locales, :out_format, :strict, :confirm]
+            args: [:locales, :out_format, :strict, :keep_order, :confirm]
 
         def remove_unused(opt = {})
           unused_keys = i18n.unused_keys(opt)
           if unused_keys.present?
             terminal_report.unused_keys(unused_keys)
             confirm_remove_unused!(unused_keys, opt)
+            if opt[:'keep-order']
```
glebm: The opt hash key would be `:keep_order`

PR #164, opened 2015-08-06
using %{key} in add_missing interpolations

stijnster:
Hi there,

thanks for your awesome work on this gem.

When using it, we noticed that using the "human_key" is not always optimal for us. So I've added the ability to add a "%{key}" value which will add the node's full_key in the interpolated value.

I've added tests for this new feature and they all turn green.

Please consider this feature to add to the main repository.

Kind regards,

Stijn


stijnster:
Of course, if you add this options to the main repo, I'll be happy to extend the current documentation ;)

glebm:
@stijnster Thank you for your contribution!  :+1:

glebm:
Release in v0.8.7! Docs updated

stijnster:
Awseome stuff! Thanks!!

PR #173, opened 2015-10-14
Adding AST as a runtime dependency

### LINKED ISSUE #172: 0.9.0.rc2 throws AST error
It appears AST is not listed as a dependency in the gemfile. 

i18n-tasks-0.9.0.rc2/lib/i18n/tasks/scanners/ruby_ast_call_finder.rb:5:in `<class:RubyAstCallFinder>': uninitialized constant AST::Processor::Mixin

rposborne:
Fixes #172


glebm:
Thanks! This was caused by whitequark/parser dependency on AST which is specified as >= 1.1, but we need v2+. Addressed in https://github.com/glebm/i18n-tasks/commit/7b62831160b6626575b30c04bcde09aa6eea09e6.

By the way, is 0.9-dev working well for you?

PR #179, opened 2015-11-19
Added note to README to stop anyone else wasting an hour of their time.

moggyboy:
Please advise if this note is incorrect; it is based on my experience only.


glebm:
That is right, though it can be easily monkey-patched if you need to [here](https://github.com/glebm/i18n-tasks/blob/7e6418e97458d3fb3d196aa3661fd5755109918f/lib/i18n/tasks/key_pattern_matching.rb#L36) by replacing the `:`s with `*`.

PR #208, opened 2016-11-01
Adds ':' to VALID_KEY_CHARS for ruby_key_literals.

### LINKED ISSUE #207: Valid keys slipping through as invalid and not showing up missing
We had an issue where we had an invalid key used in our views, but we didn't notice until it went to production because it raises a TranslationMissingError.

```yaml
# Translations
en:
  parent:
    original_value: 'Some text'
```

```ruby
# Usage in HAML

%p= t('parent.original_value:')
```

In this instance, with `strict: true`, the key is seen as [invalid on this line](https://github.com/glebm/i18n-tasks/blob/v0.9.5/lib/i18n/tasks/scanners/pattern_scanner.rb#L34).  Because of that, it's never processed. I'm not sure how you would want to handle this. Maybe providing a list of invalid keys and their usage in the application? Letting them go through and be used as missing translation?

As far as I can tell though, I18n allows this kind of key in my testing...

```yaml
# sample other translation
en:
  parent:
    'some_key:': 'A super special value'
```

```ruby
# This evalutes
%p= t('parent.some_key:')
```

jon2992:
Aims to fix #207 

Passing build: https://travis-ci.org/jon2992/i18n-tasks/builds/172334674

[This line in i18n_tasks_output_matcher](https://github.com/glebm/i18n-tasks/compare/master...jon2992:allow-colon-in-key?expand=1#diff-a633a3a22ee3798d2192bf9cfe719725R18) was adjusted to allow non-globbing keys to print out correctly if they end in a colon. I left a comment explaining my changes in `capture_std.rb`. Maybe it's just my system, but Travis also failed for that reason. It made sense to met that the tests should support ActiveSupport 4 and 5 given the support of the gem itself is that.

### DIFF
--- a/spec/support/capture_std.rb
+++ b/spec/support/capture_std.rb
@@ -1,6 +1,22 @@
 # frozen_string_literal: true
-require 'active_support/core_ext/kernel/reporting'
+
+# This is necessary because Rails 5 removed the Kernel extension of
+# that added `#silence_stream` and moved it to their testing set of
+# libraries. Therefore, I'm including this here. Technically any
+# testing will conceivably install version 5 meaning the include is
+# necessary. However, this allows us to clearly be compliant with
+# both rails 4 and 5 which the gemspec supports.
+require 'active_support/gem_version'
+
+if ActiveSupport::VERSION::MAJOR == 5
+  require 'active_support/testing/stream'
+else
+  require 'active_support/core_ext/kernel/reporting'
+end
+
 module CaptureStd
+  include ActiveSupport::Testing::Stream if defined?(ActiveSupport::Testing::Stream)
+
   def capture_stderr
     err, $stderr = $stderr, StringIO.new
     yield
--- a/spec/support/capture_std.rb
+++ b/spec/support/capture_std.rb
@@ -8,10 +8,10 @@
 # both rails 4 and 5 which the gemspec supports.
 require 'active_support/gem_version'
 
-if ActiveSupport::VERSION::MAJOR == 5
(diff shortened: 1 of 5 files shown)


jon2992:
Travis is trying to install `jruby-9.0.0.0-pre1` which works at ruby `2.2.0`. ActiveSupport requires `2.2.2`. I'm not really sure what to do about this scenario.

glebm:
I'm looking into travis

spec/support/capture_std.rb
```diff
@@ -1,6 +1,22 @@
...
+
+# This is necessary because Rails 5 removed the Kernel extension of
+# that added `#silence_stream` and moved it to their testing set of
+# libraries. Therefore, I'm including this here. Technically any
+# testing will conceivably install version 5 meaning the include is
+# necessary. However, this allows us to clearly be compliant with
+# both rails 4 and 5 which the gemspec supports.
+require 'active_support/gem_version'
+
+if ActiveSupport::VERSION::MAJOR == 5
```
glebm: I would check for version 4 instead; Then there is a chance it will also work for Rails 6

jon2992: Fair point: https://github.com/glebm/i18n-tasks/pull/208/commits/50ba5b3d5b2a4dd04e8b3983536f54f44b81dc80

glebm:
Travis seems to be having a bad day (not displaying any logs), but I'll run the tests locally and release v0.9.6 anyway.

Issue #202, opened 2016-08-26
undefined method `sample'


frodrigo:
I run i18n_tasks in a locale without provided translation (fr). But it's fails on a:

> NoMethodError: undefined method `sample' for "translation missing: fr.i18n_tasks.cmd.encourage":String
> /home/fred/.gem/gems/i18n-tasks-0.9.5/lib/i18n/tasks/reports/terminal.rb:151:in`print_success'
> /home/fred/.gem/gems/i18n-tasks-0.9.5/lib/i18n/tasks/reports/terminal.rb:55:in `unused_keys'
> /home/fred/.gem/gems/i18n-tasks-0.9.5/lib/i18n/tasks/command/options/data.rb:80:in`print_forest'
> /home/fred/.gem/gems/i18n-tasks-0.9.5/lib/i18n/tasks/command/commands/usages.rb:29:in `unused'
> /home/fred/.gem/gems/i18n-tasks-0.9.5/lib/i18n/tasks/command/commands/health.rb:20:in`health'
> /home/fred/.gem/gems/i18n-tasks-0.9.5/lib/i18n/tasks/command/commander.rb:26:in `run'
> /home/fred/.gem/gems/i18n-tasks-0.9.5/lib/i18n/tasks/cli.rb:44:in`run'

I unsuccessfully locate where come from the sample method.

glebm:
This happens because i18n-tasks doesn't have its messages translated to French.
Please make sure the `internal_locale` is set to `en` in the config (or contribute a French translation: https://github.com/glebm/i18n-tasks/tree/master/config/locales).

glebm:
Can you post the new stack trace and the output of `i18n-tasks config`?

frodrigo:
```
NoMethodError: undefined method `sample' for "translation missing: fr.i18n_tasks.cmd.encourage":String
/home/fred/.gem/gems/i18n-tasks-0.9.8/lib/i18n/tasks/reports/terminal.rb:160:in `print_success'
/home/fred/.gem/gems/i18n-tasks-0.9.8/lib/i18n/tasks/reports/terminal.rb:55:in `unused_keys'
/home/fred/.gem/gems/i18n-tasks-0.9.8/lib/i18n/tasks/command/options/data.rb:83:in `print_forest'
/home/fred/.gem/gems/i18n-tasks-0.9.8/lib/i18n/tasks/command/commands/usages.rb:30:in `unused'
[...]

PR #211, opened 2016-11-22
Remove undefined method sample on String #202

frodrigo:


glebm:
`I18n.t('i18n_tasks.cmd.encourage')` is not a string, it's an array

glebm:
Ah, this must be related to https://github.com/glebm/i18n-tasks/issues/202. Fixed in v0.9.7

frodrigo:
In fact no. The problem is still here in v0.9.8.

PR #226, opened 2017-02-03
Add scanner for model documentation as generated by annotate gem.

derari:
Some rails components, such as the form builder, automatically access i18n keys for model attributes. I18n-tasks can not automatically detect this and reports unused keys.

The [annotate gem](https://github.com/ctran/annotate_models) generates documentation for model classes that list all model fields as defined by the schema file. 

This scanner parses the field list to create key occurrences so that i18n-tasks can automatically and correctly validate the existence of model attribute keys.


glebm:
This should be a plugin, as the annotate gem is not a part of the i18n gem.

lib/i18n/tasks/scanners/model_scanner.rb
```diff
@@ -0,0 +1,38 @@
...
+    
+    MODEL_NAME = /[^\/]+(?=\.rb)/
+    
+    def scan_file(path)
+      return [] unless path =~ /.*\/models\/.*.rb/
+      model = path[MODEL_NAME]
+      text = read_file(path)
+      @patterns.flat_map do |pattern, key|
+        result = []
+        text.scan(pattern) do |_|
```
glebm: Perhaps this should scan only the beginning of the file?

lib/i18n/tasks/scanners/model_scanner.rb
```diff
@@ -0,0 +1,38 @@
...
+      super
+      @patterns = [
+        [/\#  (?<a>[\w_]+)/, 'activerecord.attributes.%{model}.%{a}']
+      ]
+    end
+    
+    MODEL_NAME = /[^\/]+(?=\.rb)/
+    
+    def scan_file(path)
+      return [] unless path =~ /.*\/models\/.*.rb/
```
glebm: Path matching should be handled during plugin registration for better performance.

PR #235, opened 2017-03-29
Use rainbow instead of GPL licensed term-ansicolor

### LINKED ISSUE #234: This library has GPL dependency
This library depends on term-ansicolor.

In Gemfile:

```ruby
s.add_dependency 'term-ansicolor', '>= 1.3.2'
```

In code:

```ruby
def warn_deprecated(message)
  log_stderr Term::ANSIColor.yellow Term::ANSIColor.bold "#{program_name}: [DEPRECATED] #{message}"
end
```

And term-ansicolor is released under GPLv2.

It means this library should be licensed under GPL or drop dependency to term-ansicolor.
https://www.gnu.org/licenses/gpl-faq.en.html#MereAggregation

ypresto:
Fixes #234

### DIFF
--- a/i18n-tasks.gemspec
+++ b/i18n-tasks.gemspec
@@ -41,7 +41,7 @@ TEXT
   s.add_dependency 'highline', '>= 1.7.3'
   s.add_dependency 'i18n'
   s.add_dependency 'parser', '>= 2.2.3.0'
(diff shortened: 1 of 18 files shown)


glebm:
Thanks! I'll fix rubocop and the failing test and release a new version.

prrrnd:
❤️

Issue #210, opened 2016-11-17
Version conflict of highline gem with Solidus


timrossinfo:
The version of the `highline` gem (`>= 1.7.3`) used in this project conflicts with a different version `~> 1.6.18`) of the same gem defined in [Solidus Core](https://github.com/solidusio/solidus/blob/master/core/solidus_core.gemspec#L30).

I have removed the strict versioning of the `highline` gem on [my fork](https://github.com/LGWgo/i18n-tasks/commit/1c676ff1f919b5df0067542db1bc44407f2f7f96). The tests still pass and the library appears to work with the older version of the gem. Can we either remove the strict versioning, or allow older versions of this dependency?

glebm:
It's fine to change it to `>= 1.6.18`, but why not bump the solidus dependency instead?

timrossinfo:
Yes, that's a good point. I thought it might be better to enable `i18n-tasks` to work with older versions of the gem (unless a newer version is required), but 4 years is quite old! I will try bumping the version in Solidus and submitting a pull request.

PR #245, opened 2017-05-22
Update readline dependency to >= 1.6.18

alepore:
Following the discussion on https://github.com/glebm/i18n-tasks/issues/210 i'd like to lower the readline required version.
I'm doing this because i want to include some i18n-tasks tasks on `solidus_i18n` and to support all the Solidus versions. 
Some of them are pretty old and had a hard dependency on readline 1.6 for some weird reasons.

Thank you ❤️

### DIFF
--- a/i18n-tasks.gemspec
+++ b/i18n-tasks.gemspec
@@ -39,7 +39,7 @@ TEXT
   s.add_dependency 'ast', '>= 2.1.0'
   s.add_dependency 'easy_translate', '>= 0.5.0'
   s.add_dependency 'erubis'
-  s.add_dependency 'highline', '>= 1.7.3'
+  s.add_dependency 'highline', '>= 1.6.18'
   s.add_dependency 'i18n'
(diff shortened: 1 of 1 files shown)


glebm:
i18n-tasks does not work well with readline 1.6
Please see the previous discussion on this matter for more information: https://github.com/glebm/i18n-tasks/issues/210

glebm:
Sorry, I've missed that you've already read the discussion in #210.
Because i18n-tasks is broken with readline 1.6, I do not want to lower the requirement (or else I'll be getting irrelevant bug reports).

If you're running it as a development tool on `solidus_i18n`, perhaps you could fork i18n-tasks and use your fork in that project, until the older versions of solidus are no longer supported.

Please do *not* support versions of solidus that no longer receive security updates, that's not good for the community.

alepore:
I've missed the incompatibility thing but i solved it anyway, thanks!

PR #251, opened 2017-07-24
Switch from erubius to erubi. Fixes #247

### LINKED ISSUE #247: Migrate to Erubi
Erubis is no longer maintained and Erubi is a mostly compatible replacement.

jcoyne:


i18n-tasks.gemspec
```diff
@@ -38,7 +38,7 @@ TEXT
   s.add_dependency 'activesupport', '>= 4.0.2'
   s.add_dependency 'ast', '>= 2.1.0'
   s.add_dependency 'easy_translate', '>= 0.5.0'
-  s.add_dependency 'erubis'
+  s.add_dependency 'erubi'
```
PikachuEXE: Should we limit the version of `erubi`?
Like `>= 1.0.0` or `~> 1.0`, like what these project do
- [rails](https://github.com/rails/rails/pull/27757/files#diff-eeca8f49241a2220a7446177d9f1c919)
- [better_error](https://github.com/charliesome/better_errors/pull/390/files#diff-bb262a96ce72496f61bd45636edbc2a9)

bartoszkopinski:
Just FYI, I think this change causes HAML to fail under Rails < 5, since it's still uses `erubis` and there is a compatibility issue:
https://github.com/haml/haml/issues/946

glebm:
This is a haml big and should be fixed on their end

glebm:
No, it is definitely not a breaking change.
It is a change to a transitive dependency. The broken code is in HAML as it relies on transitive dependencies, that no-one should ever do.

Issue #213, opened 2016-12-01
Keys with dynamic scope arguments in views are parsed as if there were no scope argument


shaneog:
Firstly, sorry for opening a support issue here!

Our app uses translations using the explicitly scoped, symbols form, e.g.
```ruby
I18n.t(:button_text, scope: [:search, :header_form])
```
We have a lot of this code in our large codebase.

We also use some dynamic keys, e.g.
```ruby
I18n.t(:title_no_results, scope: [:search, params[:action]], query: params[:q])
```

I've tried using the following hint using `i18n-tasks-use` but it doesn't work using either of the below comments on the preceding line
```haml
/ i18n-tasks-use t('search.%{params[:action]}. title_no_results')
/ i18n-tasks-use t('search.%{params[:action]}. title_no_results', query: params[:q])
```

So right now I can't use the `missing` command or RSpec test because of false positives.
Am I doing something wrong or is this type of dynamic key not supported?

glebm:
Dynamic keys need to be listed explicitly, so e.g. if the possible actions are `foo` and `bar`, the above should be:

```slim
/ i18n-tasks-use t('search.foo.title')
/ i18n-tasks-use t('search.bar.title')
```

shaneog:
Thank you for the quick response.

> In the view files, there is an additional restriction that scope must be the last argument.

Since this is used in a view, does that mean that it is not supported?
```haml
= t(:title_no_results, scope: [:search, params[:action]], query: params[:q])
```
I'll will use the explicit keys in the `i18n-tasks-use` comment, and report back.

shaneog:
Ok, I've tried using the below and it didn't work.

```haml
  / i18n-tasks-use t('search.foo.title_no_results')
  / i18n-tasks-use t('search.bar.title_no_results')
  / i18n-tasks-use t('search.foobar.title_no_results')
  / i18n-tasks-use t('search.barfoo.title_no_results')
  %h2.no-results-title= t(:title_no_results, query: params[:q], scope: [:search, params[:action]])
```

I get the following output when I run `i18n-tasks missing`:
```sh
Missing translations (1) | i18n-tasks v0.9.6
+--------+-------------------------+-------------------------------------------+
| Locale | Key                     | Value in other locales or source          |
+--------+-------------------------+-------------------------------------------+
|  all   | title_no_results_html   | app/views/search/_no_results.html.haml:10 |
+--------+-------------------------+-------------------------------------------+

```

Any ideas?

[...]

PR #255, opened 2017-08-13
Fixed pattern with key as variable for pattern with scope scanner

faizalzakaria:
Hi @glebm ,

I hope this find you well.

This https://github.com/glebm/i18n-tasks/commit/8e6b327d169901f689d56b9e79748d0bea236ebe broke our build.

With that changes, you are breaking the scenario below:

```
t(variable, scope: 'scope')
``` 

This will support both world, 

- https://github.com/glebm/i18n-tasks/issues/213
- And key as variable

Please have a look.

Thanks.


glebm:
Thanks!

PR #258, opened 2017-08-16
Remove explicit ast dependency

5t111111:
Since `parser` gem specifies `ast` as a dependency in its gemspec, you do not have to add your own `ast` dependence in a gemspec of `i18n-tasks` (like Rubocop has only `parser` in its gemspec dependencies) .

I think this change reduces dependence complexity and make maintenance easier. How does this sound to you?


glebm:
We use AST directly. Transitive dependencies should be explicit

PR #265, opened 2017-09-24
Locale data from dependencies: `data.external`

### LINKED ISSUE #264: Special handling for locale data from dependencies: `data.external`
Locale data from dependencies should not be considered unused.
Also, we should not attempt to ever modify that data (e.g. during `normalize`).

For this, a new `data.external` key should be introduced and handled accordingly with the above.

+@vsakarov

glebm:
Adds a new configuration setting, `data.external`, for locale data from dependencies.

Locale data from dependencies is never considered unused and we never
modify it.

Fixes #264 

@vsakarov PTAL

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 ## v0.9.19 (not yet released)
 
+Adds a new configuration setting, `data.external`, for locale data from external dependencies (e.g. gems).
+This locale data is never considered unused, and is never modified by i18n-tasks.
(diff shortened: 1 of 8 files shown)


vsakarov:
I like the solution, but I think is better to rename it to `external_read`. As there are two major things you can do with external gems: read locale data and search for usages (future enhancement).

glebm:
The search for usages is configured separately under `search` (e.g. this would be `search.external`), so there is no conflict.

PR #268, opened 2017-10-24
Change rainbow version lock in gemspec from ~>2.2 to ~>2.2.2

betasve:
The new Rubocop version (`0.51.0`) requires `rainbow (>= 2.2.2, < 3.0)`. 
When trying to use the new rubocop in application Bundler is complaining about incompatibility in versions
`i18n-tasks` requiring `~> 2.2` versus `rubocop` requiring `>= 2.2.2`

### DIFF
--- a/i18n-tasks.gemspec
+++ b/i18n-tasks.gemspec
@@ -42,7 +42,7 @@ TEXT
   s.add_dependency 'highline', '>= 1.7.3'
   s.add_dependency 'i18n'
   s.add_dependency 'parser', '>= 2.2.3.0'
-  s.add_dependency 'rainbow', '~> 2.2'
+  s.add_dependency 'rainbow', '~> 2.2.2'
   s.add_dependency 'terminal-table', '>= 1.5.1'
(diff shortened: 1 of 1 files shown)


glebm:
That doesn't sound right, what is the exact error message you're seeing?

betasve:
```
Bundler could not find compatible versions for gem "rainbow":
  In snapshot (Gemfile.lock):
    rainbow (= 2.2.1)

  In Gemfile:
    i18n-tasks was resolved to 0.9.18, which depends on
      rainbow (~> 2.2)

    rubocop (~> 0.51.0) was resolved to 0.51.0, which depends on
      rainbow (< 3.0, >= 2.2.2)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

It happens when I'm trying to lock `rubocop, '~> 0.51.0'` in my application and then `bundle`

betasve:
Ok, I got it... actually `bundle update` fixes the issue (nevertheless it breaks the app with other gems' versions). Will fix that issue on my end.

PR #275, opened 2018-01-16
Update the easy_translate to v0.5.1

elim:
It has:
* Add license to gemspec (MIT)
* Update googleapi hostname to support premium (NMT)
* Remove json gem dependency

Details:
https://github.com/seejohnrun/easy_translate/compare/v0.5.0...v0.5.1

* * *

I particularly want to get rid the dependency on JSON gem.

> All modern Rubies ship JSON as part of stdlib. Using the gem actually hurts multi-platform support due to build difficulties on Windows.

Cite from: <https://github.com/rails/rails/commit/f3433f7c>


elim:
Thanks!!

PR #277, opened 2018-02-28
[ci skip] Using "bundle show" in README

yhirano55:
Use `$(bundle show i18n-tasks)` instead of `$(i18n-tasks gem-path)`.
This is same as `external` section in configuration file.


glebm:
`i18n-tasks gem-path` work even if you install the gem directly, without using bundler.

PR #278, opened 2018-02-28
travis: Test on Ruby 2.5.0 (was 2.2.3)

yhirano55:


yhirano55:
Failed on rbx-3

```
Failures:
  1) I18n files are normalized FIXED
     Expected pending 'Travis CI has an older version of libyaml where the formatting differs.' to fail. No error was raised.
     # ./spec/i18n_spec.rb:20
Finished in 1 minute 14.8 seconds (files took 10.13 seconds to load)
155 examples, 1 failure, 2 pending
Failed examples:
rspec ./spec/i18n_spec.rb:20 # I18n files are normalized
```

glebm:
We test on the minimum supported version to make sure we don't accidentally use Ruby features that are not compatible with older versions of Ruby.

I'll fix the failing rbx test separately.

yhirano55:
I see. Thanks.

PR #294, opened 2018-07-30
add DeepL Pro artificial intelligence translation service

neumayr:
Hey @glebm 👋

* add [DeepL Pro](https://www.deepl.com/pro) artificial intelligence translation service ✨
  * DeepL trains artificial intelligence to understand and translate texts.
  * 🇬🇧 🇺🇸 🇩🇪 🇫🇷 🇪🇸 🇮🇹 🇳🇱 🇵🇱
  * [TechCrunch: DeepL schools other online translators with clever machine learning](https://techcrunch.com/2017/08/29/deepl-schools-other-online-translators-with-clever-machine-learning/)
* use [`deepl-rb`](https://github.com/wikiti/deepl-rb) gem for query the API 💎
  * Kudos to @wikiti for helping with the latest `v2.1.0` release that adds `ignore_tags` 💪
* currently the cmd is `i18n-tasks translate-missing-deepl` because I wasn’t sure how you would structure the commands with multiple vendors. What do you think?

---
👷‍♂️ I have a `DEEPL_AUTH_KEY` I can share for testing and development
⚠️ I had to prefix some private methods with `deepl_` (like `deepl_to_values`) because I noticed the code was using the implementation from `GoogleTranslation`. 🧐 Where I didn't get behind it. I hope we can collaborate to iron out all the issues. ✌️

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,7 +367,7 @@ Put the key in `GOOGLE_TRANSLATE_API_KEY` environment variable or in the config 
 ```yaml
 # config/i18n-tasks.yml
 translation:
-  api_key: <Google Translate API key>
+  google_translate_api_key: <Google Translate API key>
 ```
 
 <a name="deepl-translation-config"></a>
@@ -378,7 +378,7 @@ translation:
 ```yaml
 # config/i18n-tasks.yml
 translation:
-  api_key: <Deep Pro API key>
+  deepl_api_key: <Deep Pro API key>
 ```
 
 ## Interactive console
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ i18n-tasks helps you find and manage missing and unused translations.
 This gem analyses code statically for key usages, such as `I18n.t('some.key')`, in order to:
 
 * Report keys that are missing or unused.
-* Pre-fill missing keys, optionally from Google Translate.
+* Pre-fill missing keys, optionally from Google Translate or DeepL Pro.
 * Remove unused keys.
 
 Thus addressing the two main problems of [i18n gem][i18n-gem] design:
@@ -83,12 +83,22 @@ Usage: i18n-tasks add-missing [options] [locale ...]
 
 ### Google Translate missing keys
 
-Translate missing values with Google Translate ([more below on the API key](#translation-config)).
+Translate missing values with Google Translate ([more below on the API key](#google-translation-config)).
 
 ```console
 $ i18n-tasks translate-missing
 # accepts from and locales options:
 $ i18n-tasks translate-missing --from base es fr
+```
+
+### DeepL Pro Translate missing keys
+
+Translate missing values with DeepL Pro Translate ([more below on the API key](#deepl-translation-config)).
+
+```console
+$ i18n-tasks translate-missing
+# accepts from and locales options:
+$ i18n-tasks translate-missing-deepl --from en
 ```
 
 ### Find usages
@@ -295,7 +305,7 @@ data:
     - 'config/locales/%{locale}.yml'
 ```
 
-If you want to have i18n-tasks reorganize your existing keys using `data.write`, either set the router to 
+If you want to have i18n-tasks reorganize your existing keys using `data.write`, either set the router to
 `pattern_router` as above, or run `i18n-tasks normalize -p` (forcing the use of the pattern router for that run).
 
 ##### Key pattern syntax
@@ -340,7 +350,7 @@ For more complex cases, you can implement a [custom scanner][custom-scanner-docs
 
 See the [config file][config] to find out more.
 
-<a name="translation-config"></a>
+<a name="google-translation-config"></a>
 ### Google Translate
 
 `i18n-tasks translate-missing` requires a Google Translate API key, get it at [Google API Console](https://code.google.com/apis/console).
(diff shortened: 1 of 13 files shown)


lib/i18n/tasks/deepl_translation.rb
```diff
@@ -0,0 +1,125 @@
...
+      result.concat(reference_key_vals)
+      result.sort! { |a, b| key_pos[a[0]] <=> key_pos[b[0]] }
+      result
+    end
+
+    # @param [Array<[String, Object]>] list of key-value pairs
+    # @return [Array<[String, Object]>] translated list
+    def fetch_translations(list, opts)
+      options = {
+        tag_handling: 'xml',
```
glebm: Should it have `tag_handling: 'xml'` for non-HTML keys?

README.md
```diff
@@ -360,6 +370,17 @@ translation:
...
 
+<a name="deepl-translation-config"></a>
+### DeepL Pro Translate
+
+`i18n-tasks translate-missing` requires a DeepL Pro API key, get it at [DeepL](https://www.deepl.com/pro).
+
+```yaml
+# config/i18n-tasks.yml
+translation:
+  api_key: <Deep Pro API key>
```
glebm: Let's have separate settings for different API keys. `google_translate_api_key` and `deepl_api_key`. (`api_key` can then work as `google_translate_api_key` with a deprecation warning`)

lib/i18n/tasks/command/commands/missing.rb
```diff
@@ -46,6 +47,20 @@ def translate_missing(opt = {})
           print_forest translated, opt
         end
 
+        # DeepL Pro Translate
+        cmd :translate_missing_deepl,
```
glebm: Let's add a `--provider` flag to `translate_missing` instead of having a separate command that is otherwise identical.

lib/i18n/tasks/deepl_translation.rb
```diff
@@ -0,0 +1,125 @@
...
+    def deepl_translate_forest(forest, from)
+      forest.inject empty_forest do |result, root|
+        translated = translate_list(root.key_values(root: true), to: root.key, from: from)
+        result.merge! Data::Tree::Siblings.from_flat_pairs(translated)
+      end
+    end
+
+    # @param [Array<[String, Object]>] list of key-value pairs
+    # @return [Array<[String, Object]>] translated list
+    def translate_list(list, opts) # rubocop:disable Metrics/AbcSize
```
glebm: Consider prefixing all methods here and in `google_translation.rb` with `google_` / `deepl_` and extracting common methods into a `base_translation.rb` module.

glebm:
Thanks @neumayr, I've left some comments.

lib/i18n/tasks/command/commands/missing.rb
neumayr: Could you check if `Command::Options::Locales` could be good as well?

lib/i18n/tasks/deepl_translation.rb
neumayr: I prefixed the methods with `google_` / `deepl_`. The only method is `*_to_values(list)` that would identify for a `base_translation.rb` Do you think it's worth sharing?

lib/i18n/tasks/deepl_translation.rb
neumayr: thx 👍

lib/i18n/tasks/deepl_translation.rb
```diff
@@ -0,0 +1,124 @@
...
+      result.concat(reference_key_vals)
+      result.sort! { |a, b| key_pos[a[0]] <=> key_pos[b[0]] }
+      result
+    end
+
+    # @param [Array<[String, Object]>] list of key-value pairs
+    # @return [Array<[String, Object]>] translated list
+    def fetch_translations(list, opts)
+      options = {
+        ignore_tags: %w[i18n]
```
glebm: Does `ignore_tags` work even without `tag_handling: 'xml'`? 
If not, we may need to use the same trick as in the Google Translate implementation

lib/i18n/tasks/deepl_translation.rb
glebm: Not worth it unless we re-structure it to share more methods. I can look into it after merging, thanks.

lib/i18n/tasks/deepl_translation.rb
neumayr: Yes, `ignore_tags` works even without other params. And should be fixed, because `%{user}` interpolation can be with `_html` / `.html` keys and on simple text tags.

📝 https://www.deepl.com/api.html

glebm:
Thanks, merged! Can you email me the key so I may add it to Travis?

glebm:
I've moved the shared code to a base class in https://github.com/glebm/i18n-tasks/commit/d82679da814869f5d7693174a2184f4c7cb2feaf

dgilperez:
Hi! 

After upgrading the gem, I just noticed this new dependency on our project. 

While this service looks great, I'd like to suggest to implement it as an optional dependency somehow - adding this gem to any project using `i18n-tasks`, regardless of they use Deepl or not (most won't). 

Probably this will involve removing de dependency from the `gemspec` and adding a note to README, with the need to add `gem 'deepl'` to the `Gemfile`, right? 

Would you accept a PR?

PR #298, opened 2018-08-26
Test against the latest ruby

deivid-rodriguez:
Maybe you should actually add all supported major rubies in here?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -52,19 +52,7 @@ Style/FormatStringToken:
 Style/GuardClause:
   Enabled: false
 
-Layout/IndentHeredoc:
-  # TODO(glebm): Remove this when we update the minimum Ruby version to 2.3+.
-  Enabled: false
-
 Style/MultilineBlockChain:
-  Enabled: false
-
-Style/NumericPredicate:
-  # TODO(glebm): Remove this when we update the minimum Ruby version to 2.3+.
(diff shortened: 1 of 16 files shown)


deivid-rodriguez:
I also observed that you use `TargetRubyVersion: 2.3` in `.rubocop.yml`, but `required_ruby_version = '~> 2.1'` in gemspec. This is dangerous because you might accidentally break support by using newer rubies features, specially since you don't test all versions. You should probably reconcile that too.

deivid-rodriguez:
Any thoughts here @glebm? I'll gladly make the extra changes I commented if you agree with them!

glebm:
Sounds good to me and let's remove support for older rubies. https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

deivid-rodriguez:
Absolutely, let me add those changes!

deivid-rodriguez:
It's ready, have a look! I also fixed some TODO's were meant to be fixed after dropping 2.2 support. I fully trusted rubocop auto-correction here (plus manually fixed the non-autocorrectable `Style/NumericPredicate` offenses. I looked at the generated diff and looks correct to me?

deivid-rodriguez:
Also, I didn't touch the changelog because I noticed that you normally update it at relase time.

PR #301, opened 2018-09-28
Command eq-base returns an error if any key is having the same value.

pcorpet:
This would allow us to make this a blocking part of our CI.


D-system:
Also looking forward for this PR.

PR #302, opened 2018-10-04
Add MRI ruby version requirement for 9.0

joeltaylor:
The introduction of [whitequark/parser](https://github.com/whitequark/parser) requires the latest minor patch of MRI Ruby and will raise errors. Figured it'd be worth noting for those who aren't able to upgrade so they know to stick to `0.8.7` until they can upgrade.


glebm:
I haven't actually encountered any issues with using the non-latest patch versions of MRI.

> will raise errors. 

It just prints warnings for me. Pretty sure these are safe to ignore. Way better than using 0.8.x anyhow, because a lot of bugs were fixed since then.

joeltaylor:
That was a bad choice of words on my part. While upgrading, I thought it had also raised parse errors in additions to those warnings, but I've been unable to replicate (which is great!)

PR #304, opened 2018-10-17
Add check-consistent-interpolations

### LINKED ISSUE #303: Detect misused variables based on reference locale
Would this be something you'd want to have in i18n-tasks? If there is a master locale, and it uses certain variables in certain keys, ensure that all other locales use the same variables in the same keys.

Use case: We use Weblate on Mastodon to let people submit translations. Some people translated variable names, thereby breaking some of Mastodon's pages when viewed in those languages. It's very hard to detect when this happens.

Gargron:
Verify that non-base translations are using variable names defined in the base translation.

Fix #303

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -33,6 +33,7 @@ en:
             %{value_or_default_or_human_key}
       desc:
         add_missing: add missing keys to locale data
+        check_consistent_interpolations: verify that all translations use correct interpolation variables
         check_normalized: verify that all translation data is normalized
         config: display i18n-tasks configuration
         data: show locale data
@@ -102,6 +103,8 @@ en:
           Google Translate returned no results. Make sure billing information is set at https://code.google.com/apis/console.
     health:
       no_keys_detected: No keys detected. Check data.read in config/i18n-tasks.yml.
+    inconsistent_interpolation:
+      none: No inconsistent interpolations found.
     missing:
       details_title: Value in other locales or source
       none: No translations are missing.
--- a/config/locales/ru.yml
+++ b/config/locales/ru.yml
@@ -30,6 +30,8 @@ ru:
             %{value_or_default_or_human_key}
       desc:
         add_missing: добавить недостающие ключи к переводам
+        check_consistent_interpolations: убедитесь, что во всех переводах используются правильные
+          интерполяционные переменные
         check_normalized: проверить, что все файлы переводов нормализованы
         config: показать конфигурацию
         data: показать данные переводов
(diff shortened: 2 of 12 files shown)


Gargron:
@glebm I was not able to figure out why `check-consistent-interpolations` does not appear in the list when `bin/i18n-tasks` is executed, and I was not able to call it through that. Even if I renamed the health command to health_test, the output remained the same, which suggests to me that my workflow is incorrect (I do not develop CLI gems often)

glebm:
Does `bundle exec i18n-tasks` work for you?

lib/i18n/tasks/wrong_interpolations.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      compare_variables = node.value.scan(VARIABLE_REGEX)
+
+      return if same_array?(base_variables, compare_variables)
+
+      tree.merge!(select_key(key, locale))
+    end
+
+    def select_key(key, locale)
+      data[locale].select_keys(root: false) { |search_key| search_key == key }
```
glebm: This is rather inefficient as it traverses the whole tree. Consider something like this instead:

```ruby
tree.append!(
  node.walk_to_root.reduce(nil) { |p, c| [p.derive(children: c)] }.tap do |root|
    root.data[:type] = :inconsistent_interpolation
  end
)
```

lib/i18n/tasks/wrong_interpolations.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+      end
+
+      tree
+    end
+
+    private
+
+    def check_in_other_locale(tree, key, base_variables, locale)
+      node = data[locale].first.children[key]
```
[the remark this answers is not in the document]
Gargron: `data[locale][key]` does not do the same thing as `data[locale].first.children[key]`, I have picked this pattern from `equal_values_tree` in `missing_keys.rb`

lib/i18n/tasks/wrong_interpolations.rb
Gargron: That code does not work as is, and I do not understand walk_to_root/derive enough to figure out why. Won't append also fail when multiple nodes come from the same locale e.g. `es`?

Gargron:
OK, `bundle exec` worked :grin: 

I was not able to use the faster traversal method, and I was not able to simplify the one method below the rubocop threshold, but otherwise it works now.

glebm:
Released in v0.9.26! I've renamed some methods and the ignore setting before releasing.

PR #307, opened 2018-10-23
Eq-base: returned value

### LINKED ISSUE #306: "eq-base" wrong return value
With or without similar value as base, the returned value is the same.
```
> bundle exec i18n-tasks eq-base
Same value as en (1) | i18n-tasks v0.9.25
+--------+-----------+-------+
| Locale | Key       | Value |
+--------+-----------+-------+
|   ja   | fund.isin | ISIN  |
+--------+-----------+-------+
> echo $?
0
> bundle exec i18n-tasks eq-base
No translations are the same as base value
> echo $?
0
```

Expected value: 0 when empty and 1 when not empty.

D-system:
Return 1 on error/empty list.

Fix #306 

```
> bundle exec i18n-tasks eq-base
Same value as en (1) | i18n-tasks v0.9.25
+--------+-----------+-------+
| Locale | Key       | Value |
+--------+-----------+-------+
|   ja   | fund.isin | ISIN  |
+--------+-----------+-------+
> echo $?
1
> bundle exec i18n-tasks eq-base
No translations are the same as base value
> echo $?
0
```

### DIFF
--- a/lib/i18n/tasks/command/commands/eq_base.rb
+++ b/lib/i18n/tasks/command/commands/eq_base.rb
@@ -12,7 +12,9 @@ module I18n::Tasks
             args: %i[locales out_format]
 
         def eq_base(opt = {})
-          print_forest i18n.eq_base_keys(opt), opt, :eq_base_keys
+          forest = i18n.eq_base_keys(opt)
+          print_forest forest, opt, :eq_base_keys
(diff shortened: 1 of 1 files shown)


D-system:
Duplicates https://github.com/glebm/i18n-tasks/pull/301

PR #309, opened 2018-10-24
Add missing-plural-keys

### LINKED ISSUE #308: Detect incomplete pluralizations
In languages like Arabic, a lot of keys are required for using pluralization: not simply "one" and "other", but "zero", "one", "two", "few" and "many", and there are many languages like that. Having missing keys there breaks the whole page. It would be nice to have another diagnostic command that would check locale files for presence of pluralized keys, and ensure they have all the pluralizations necessary for that locale.

Gargron:
Fix #308

### DIFF
--- a/lib/i18n/tasks/missing_keys.rb
+++ b/lib/i18n/tasks/missing_keys.rb
@@ -2,12 +2,12 @@
 
 require 'set'
 module I18n::Tasks
-  module MissingKeys
-    MISSING_TYPES = %w(
+  module MissingKeys # rubocop:disable Metrics/ModuleLength
+    MISSING_TYPES = %w[
       used
       diff
       plural
-    ).freeze
+    ].freeze
 
     def self.missing_keys_types
       @missing_keys_types ||= MISSING_TYPES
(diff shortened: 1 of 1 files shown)


Gargron:
@glebm So here's the deal. If you require `i18n/tasks` from a Rails app that has `rails-i18n` as a dependency, then the method works. But from within the standalone executable, it does not work. I cannot figure out how to include pluralization data from rails-i18n.

But other than that, the code works.

glebm:
>  I cannot figure out how to include pluralization data from rails-i18n into the gem itself

This gem doesn't and shouldn't depend on rails-i18n. Plural keys are defined in the gem itself:
https://github.com/glebm/i18n-tasks/blob/master/lib/i18n/tasks/plural_keys.rb

If you want to include translations from another gem into i18n-tasks analysis, use `data.external` in `config/i18n-tasks.yml` (see readme and the config template).

Gargron:
There are some rubocop offences I can't really fix

glebm:
For `Metrics/AbcSize`, feel free to disable it where it makes sense (see https://github.com/rubocop-hq/rubocop/blob/master/manual/configuration.md#disabling-cops-within-source-code)

Gargron:
Done!

It's now part of `missing`. I have also removed some dead code that was not called from anywhere, such as glyph/summary definitions for the missing types, and the glyph renderer in the terminal report.

Also, rails-i18n does not call i18n-tasks code, only the stand-alone executable, so I don't think it will be a problem.

I do not have the bandwidth to implement an `add-missing` behaviour here. My guess is just that if the translations for all plural keys are identical, the translation should be turned from a tree to a single leaf. It would be convenient but I don't really need it right now.

glebm:
Thank you! This and interpolation checking are great additions!

PR #312, opened 2018-10-26
Do not consider a plural node if it only has "other" with no count

### LINKED ISSUE #310: Missing plural keys wrong locale displayed, does not respect ignore list
This is my mistake. For some reason, a missing plural in "it" shows up as missing in "zh-TW", and before that, missing plurals in "cs" also showed up as "zh-TW". I have no idea why, must some key must be getting overwritten somewhere.

Also, I discovered I have a weird case where I have a structure like:

```yaml
platforms:
  other: unknown platform
```

And unfortunately it shows up as missing "one" even though it's not a plural. Maybe when deciding based on only "other" being present, a condition should be that it contains the "count" variable as well?

Gargron:
Fix #310

### DIFF
--- a/lib/i18n/tasks/plural_keys.rb
+++ b/lib/i18n/tasks/plural_keys.rb
@@ -46,7 +46,12 @@ module I18n::Tasks::PluralKeys
   end
 
   def plural_forms?(s)
+    return false if non_plural_other?(s)
     s.present? && s.all? { |node| node.leaf? && plural_suffix?(node.key) }
+  end
+
+  def non_plural_other?(s)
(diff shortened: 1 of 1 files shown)


Gargron:
Addressed

Also, it would be amazing if this could make it into a gem release quickly because I would prefer not to push out a Mastodon release with a git-based dependency if I can help it, and the release is pretty much only waiting for i18n things right now

glebm:
@Gargron I'll try to release tonight. I'm in Brussels for a conference but hopefully I'll have time.

glebm:
@Gargron Released

glebm:
i had to revert this due to https://github.com/glebm/i18n-tasks/issues/461, hopefully this doesn't cause any issues.

PR #313, opened 2018-10-27
Include file name in CommandError raised during unsuccessful file load

aleksandrs-ledovskis:
Currently, when locale file is not correctly formatted, i18n-tasks shows similar message in CLI.

```shell
$ i18n-tasks health
i18n-tasks: yaml parse error: ({}): mapping values are not allowed in this context at line 17 column 67 
```

When number of YAMLs/JSONs is large and they are potentially spread across different libraries, it's difficult to pinpoint exact source of the issue.

This change makes error messages originating from incorrectly formatted YAML/JSON files to include file name.

```shell
$ i18n-tasks health
i18n-tasks: yaml parse error: ({}): mapping values are not allowed in this context at line 17 column 67 (file: /somewhere/some-file.yml)
```


aleksandrs-ledovskis:
Build initially failed due to Rubinius' JSON [implementation](https://github.com/flori/json) able to handle/parse `{` as `{}`. Consequently, switched to literal `-` as example of invalid JSON.

PR #320, opened 2018-11-24
Update README and post installation instructions message

dportalesr:
- Recommend usage of 'bundle exec'.
- Update post installation instructions to use bundle exec.

My enviroment couldn't find the `i18n-task` executable without the use of `bundle exec`.
Showing this recommendation in the post installation message will make this more friendly for projects using this gem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ $ cp $(i18n-tasks gem-path)/templates/minitest/i18n_test.rb test/
 
 ## Usage
 
-Run `i18n-tasks` to get the list of all the tasks with short descriptions.
+Run `bundle exec i18n-tasks` to get the list of all the tasks with short descriptions.
 
 ### Check health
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ $ cp $(i18n-tasks gem-path)/templates/minitest/i18n_test.rb test/
 
 ## Usage
 
-Run `i18n-tasks` to get the list of all the tasks with short descriptions.
+Run `i18n-tasks` or `bundle exec i18n-tasks` (recommended) to get the list of all the tasks with short descriptions.
```
glebm: Let's just say `bundle exec i18n-tasks`.

dportalesr: Done.

dportalesr: The rest of the document still refers running it without the `bundle exec` but I'm not sure if it's a good idea to replace it everywhere.

glebm:
Thanks!

PR #325, opened 2019-01-18
Add example of usage for "Key pattern syntax"

WaKeMaTTa:
**Documentation**

- [x] Add example of usage for "Key pattern syntax"


glebm:
Thanks!

PR #327, opened 2019-02-19
Support pattern args in remove_unused command

michiomochi:
There are some keys that I don't want to delete in my app, but are subject to deletion by `remove_unused` command. (e.g. activerecord, devise, some gems keys, and keys being dynamically referenced 😑  )
It is useful to have this in order to exclude them from deletion.


glebm:
Thanks!

michiomochi:
寿司ゆき:thanks

Issue #335, opened 2019-11-15
Rails I18n 5.1.3 and `:'en-gb'` vs `:'en-GB'`


Foozie3Moons:
It looks like rails I18n 5.1.3 defines British English locale as `:'en-GB'` which breaks the current `required_plural_keys_for_locale(locale)!` method:

```
/Users/mabell/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/i18n-tasks-0.9.29/lib/i18n/tasks/missing_keys.rb:85:in `required_plural_keys_for_locale': undefined method `[]' for nil:NilClass (NoMethodError)
```

Stepped through to figure out that it was `:en-gb` that was not matching up properly in the hash, leading to the nil hash accessing error above. This one's a tough one to fix. But this seems to work for me:

```ruby
def required_plural_keys_for_locale(locale)
  @plural_keys_for_locale ||= {}
  return @plural_keys_for_locale[locale] if @plural_keys_for_locale.key?(locale)
  @plural_keys_for_locale[locale] = plural_keys_for_locale(locale)
end

private

def plural_keys_for_locale(locale)
  configuration = load_rails_i18n_pluralization!(locale)
  if configuration[locale.to_sym].nil?
    alternate_locale = alternate_locale_from(locale)
    return Set.new if configuration[alternate_locale].nil?
    return set_from_rails_i18n_pluralization(configuration, alternate_locale)
  end
  set_from_rails_i18n_pluralization(configuration, locale)
rescue SystemCallError, IOError
  Set.new
end

def alternate_locale_from(locale)
  re = /(\w{2})-*(\w{2})*/
  match = locale.match(re)
  language_code = match[1]
  country_code = match[2]
  country_code.upcase
  "#{language_code}-#{country_code}".freeze
end

def set_from_rails_i18n_pluralization(configuration, locale)
  Set.new(configuration[locale.to_sym][:i18n][:plural][:keys])
end
```

I'll open a PR later tonight.

davidwessman:
@glebm close by #357 ?

glebm:
Yep!

PR #336, opened 2019-11-15
Trying capitalized version of country code when accessing I18n configuration data

Foozie3Moons:
The fix described in issue https://github.com/glebm/i18n-tasks/issues/335

### DIFF
--- a/lib/i18n/tasks/missing_keys.rb
+++ b/lib/i18n/tasks/missing_keys.rb
@@ -79,12 +79,7 @@ module I18n::Tasks
     def required_plural_keys_for_locale(locale)
       @plural_keys_for_locale ||= {}
       return @plural_keys_for_locale[locale] if @plural_keys_for_locale.key?(locale)
-      @plural_keys_for_locale[locale] =
-        begin
-          Set.new(load_rails_i18n_pluralization!(locale)[locale.to_sym][:i18n][:plural][:keys])
-        rescue SystemCallError, IOError
-          Set.new
-        end
+      @plural_keys_for_locale[locale] = plural_keys_for_locale(locale)
     end
 
     # Loads rails-i18n pluralization config for the given locale.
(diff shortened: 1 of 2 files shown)


glebm:
Please add a test

Foozie3Moons:
Will do.

Foozie3Moons:
Should be good to go, is this test coverage for this satisfactory?

spec/missing_keys_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    def configuration_from(locale)
+      {
+        "#{locale}": {
+          i18n: {
+            plural:{
+              keys: [
+                :one,
+                :other
+              ],
+            rule: lambda {}
```
glebm: Is this `rule` necessary for the test?

lib/i18n/tasks/missing_keys.rb
```diff
@@ -154,5 +149,32 @@ def collapse_same_key_in_locales!(forest)
...
+    rescue SystemCallError, IOError
+      Set.new
+    end
+
+    def alternate_locale_from(locale)
+      re = /(\w{2})-*(\w{2})*/
+      match = locale.match(re)
+      language_code = match[1]
+      country_code = match[2]
+      country_code.upcase
```
glebm: This doesn't actually modify the `country_code` but the test didn't fail

Foozie3Moons: You're right, I think this solved my problem because it returned an empty set for `en-GB`.

spec/missing_keys_spec.rb
Foozie3Moons: It's not necessary for the test, but it is included in the eval from `load_rails_i18n_pluralization!(locale)`. I can remove it if you'd prefer.

lib/i18n/tasks/missing_keys.rb
Foozie3Moons: Updated the code and tests, thanks for catching that.

lib/i18n/tasks/missing_keys.rb
```diff
@@ -154,5 +150,32 @@ def collapse_same_key_in_locales!(forest)
...
+
+        return set_from_rails_i18n_pluralization(configuration, alternate_locale)
+      end
+      set_from_rails_i18n_pluralization(configuration, locale)
+    rescue SystemCallError, IOError
+      Set.new
+    end
+
+    def alternate_locale_from(locale)
+      re = /(\w{2})-*(\w{2})*/
```
glebm: The second group can be 3 letters, e.g. https://github.com/svenfuchs/rails-i18n/blob/254b1b43730cde87a727b83b569c08fc59b7e3fa/rails/locale/zh-YUE.yml

PR #341, opened 2020-01-09
Remove RainbowUtils

pocke:
`RainbowUtils` has been introduced by #235 to use `faint` method that is proposed by https://github.com/sickill/rainbow/pull/53.

The proposal has been accepted and it is contained in v2.2.2.
i18n-tasks requires rainbow gem v2.2.2 or later.
https://github.com/glebm/i18n-tasks/blob/5a41e4b6c679341b78943ab74a09a921994a1934/i18n-tasks.gemspec#L43

So we can remove the RainbowUtils module.

cc/ @ypresto


ypresto:
Wow, very old TODO comment. 😮 Thanks for cleaning it up! 🐱 :pray:

glebm:
Thanks!

PR #342, opened 2020-01-09
Suppress keyword arguments warnings in Ruby 2.7

pocke:
It suppresses keyword arguments warnings in Ruby 2.7. And it works well with Ruby 2.8 (the master branch).

# Problem

I18n-tasks displays warnings on Ruby 2.7.
I found the following warnings with `bundle exec rake`.

```
/path/to//glebm/i18n-tasks/spec/file_system_data_spec.rb:88: warning: Passing the keyword argument as the last hash parameter is deprecated
/path/to//glebm/i18n-tasks/spec/file_system_data_spec.rb:123: warning: Passing the keyword argument as the last hash parameter is deprecated
/path/to//glebm/i18n-tasks/lib/i18n/tasks/data/tree/siblings.rb:276: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/path/to//glebm/i18n-tasks/spec/support/trees.rb:20: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/path/to//glebm/i18n-tasks/lib/i18n/tasks/data/tree/traversal.rb:154: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/path/to//glebm/i18n-tasks/spec/support/keys_and_occurrences.rb:13: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

And the test fails on Ruby 2.8.

# Solution

Add `**` to keyword arguments, etc.
see https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

And this pull request also enables Ruby 2.7 on Travis CI.

### DIFF
--- a/lib/i18n/tasks/data/tree/traversal.rb
+++ b/lib/i18n/tasks/data/tree/traversal.rb
@@ -60,14 +60,12 @@ module I18n::Tasks
         self
       end
 
-      def key_names(opts = {})
-        opts[:root] = false unless opts.key?(:root)
-        keys(opts).map { |key, _node| key }
+      def key_names(root: false)
+        keys(root: root).map { |key, _node| key }
       end
 
-      def key_values(opts = {})
-        opts[:root] = false unless opts.key?(:root)
-        keys(opts).map { |key, node| [key, node.value] }
+      def key_values(root: false)
+        keys(root: root).map { |key, node| [key, node.value] }
       end
 
       def root_key_values(sort = false)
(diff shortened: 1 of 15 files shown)


lib/i18n/tasks/data/tree/traversal.rb
```diff
@@ -143,18 +141,6 @@ def get_nodes_by_key_filter(root: false, &block)
         matches
       end
 
-      # @return [Siblings]
-      def intersect_keys(other_tree, key_opts = {}, &block)
```
pocke: `intersect_keys` method is only called by test, so I removed it.

glebm: This method is public and may be used by plugins / user extensions so let's keep it

pocke: Thank you for reviewing. I overlooked the plugin system.

I restored the method and spec in 40c91279da5277079deba6bb1992189b2a4fc6d5.
And I modify the method to support Ruby 3 keyword arguments.

PR #343, opened 2020-02-27
Add Yandex translator option to translate_missing functionality

SlakrHakr:
translate_missing can be incredibly useful but the two existing options both cost money to use. Yandex Translate is a potential free option for people.

### DIFF
--- a/lib/i18n/tasks/translators/yandex_translator.rb
+++ b/lib/i18n/tasks/translators/yandex_translator.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'i18n/tasks/translators/base_translator'
+
+module I18n::Tasks::Translators
+  class YandexTranslator < BaseTranslator
+    def initialize(*)
+      begin
+        require 'yandex-translator'
+      rescue LoadError
+        raise ::I18n::Tasks::CommandError, "Add gem 'yandex-translator' to your Gemfile to use this command"
+      end
+      super
+    end
+
+    protected
+
+    def translate_values(list, **options)
+      list.map { |item| translator.translate(item, options) }
+    end
+
+    def options_for_translate_values(from:, to:, **options)
+      options.merge(
+        from: to_yandex_compatible_locale(from),
+        to: to_yandex_compatible_locale(to)
+      )
+    end
+
+    def no_results_error_message
+      I18n.t('i18n_tasks.yandex_translate.errors.no_results')
+    end
+
+    private
+
+    # Convert 'es-ES' to 'es'
+    def to_yandex_compatible_locale(locale)
+      return locale unless locale.include?('-')
+      locale.split('-', 2).first
+    end
+
+    def translator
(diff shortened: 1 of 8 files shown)


lib/i18n/tasks/version.rb
```diff
@@ -2,6 +2,6 @@
 
 module I18n
   module Tasks
-    VERSION = '0.9.30'
+    VERSION = '0.10.1'
```
SlakrHakr: I wasn't sure what version scheme made sense for this level of change and project. Happy to set to whatever is needed.

glebm: I'll bump the version in a separate commit (will be 0.9.31 instead because most users have the version locked to `~> 0.9.x`)

lib/i18n/tasks/translators/yandex_translator.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def options_for_translate_values(from:, to:, **options)
+      options.merge(
+        from: to_yandex_compatible_locale(from),
+        to: to_yandex_compatible_locale(to)
+      )
+    end
+
+    def options_for_html
+      {}
```
glebm: The Yandex API needs `format: 'html'` for translating HTML values
https://tech.yandex.com/translate/doc/dg/reference/translate-docpage/

I see that the yandex-translate gem doesn't support the `format` option yet.
We should make sure HTML keys are supported before merging this, because otherwise they will have broken markup.

Looks like there is an abandoned PR for this https://github.com/aegorov/yandex-translator/pull/13

lib/i18n/tasks/version.rb
SlakrHakr: reverted my version change
https://github.com/KicksiteDev/i18n-tasks/commit/51efc00b555409ed2447edcc7d078597ff24967d

lib/i18n/tasks/translators/yandex_translator.rb
SlakrHakr: Opened a pr to do that work over there and updated the implementation here to utilize it:
https://github.com/aegorov/yandex-translator/pull/14

lib/i18n/tasks/translators/yandex_translator.rb
```diff
@@ -0,0 +1,61 @@
...
+      rescue LoadError
+        raise ::I18n::Tasks::CommandError, "Add gem 'yandex-translator' to your Gemfile to use this command"
+      end
+      super
+    end
+
+    protected
+
+    def translate_values(list, **options)
+      list.map { |item| translator.translate(item, options) }
```
glebm: Note: This is going to cause lots of requests to the Yandex API, processed in-sequence (slow).

It would be better to pass multiple `text` parameters to the request, batched to avoid hitting the 10,000 character limit, each batch in its own Thread or Fiber. This should be implemented in the yandex gem.

PR #357, opened 2020-10-02
Trying capitalized version of country code when accessing I18n configuration data

bmedenwald:
This is the follow-on to #336 that seems to have died last year. This issue is broken for those that need country codes (like me).

I've addressed the possibility of three digit country codes (like zh-YUE) and added a test that confirms it as well.

Otherwise, the PR code is largely @Foozie3Moons


glebm:
Thanks!

PR #365, opened 2020-12-16
Check external key is present in the current locale

### LINKED ISSUE #364: How to require app to amend gem's translations with a new locale?
I'm struggling with the following:

- I'm using a (private) gem that provides translations for the `:nl` and `:en` locales
- My app needs to also support the `:de` locale
- I would like to know if I've provided all needed translations for `:de`

How can I configure i18n-tasks to make this happen?

mvz:
If the key is only checked in the base locale, the missing keys check will not detect keys that are missing for other locales in an external gem.

Fixes #364


glebm:
Thanks!

Issue #370, opened 2021-02-12
Code incompatible with ruby 3


mildred:
https://github.com/glebm/i18n-tasks/blob/1d1f4205e66a48b41dda4225f29d7806674e0249/lib/i18n/tasks/command/dsl.rb#L13-L15

I believe it should be:

```ruby
 def t(*args, **kargs) 
   I18n.t(*args, **kargs) 
 end 
```

The error: ```i18n-1.8.8/lib/i18n.rb:196:in `translate': wrong number of arguments (given 2, expected 0..1) (ArgumentError)```

MatthewKennedy:
I'm getting the same issues.

glebm:
Released v0.9.34

PR #374, opened 2021-03-01
Add ruby 3 support

artplan1:
Fixes https://github.com/glebm/i18n-tasks/issues/370
Method I used should be compatible with all rubies, more info https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/ and in https://discuss.rubyonrails.org/t/new-2-7-3-0-keyword-argument-pain-point/74980/47
example: https://github.com/thbar/kiba/pull/93

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ dist: focal
 language: ruby
 rvm:
 - ruby-head
+- 3.0.0
 - 2.7.0
 - 2.6.5
 - 2.5.7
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,6 +3,7 @@ dist: focal
 language: ruby
 rvm:
 - ruby-head
```
artplan1, line 5: head = 3 at the moment, but head is allowed to fail. not sure if I should keep 3.0.0 here?

glebm:
Thanks!

Did this instead because we know that the `t` method accepts keyword arguments https://github.com/glebm/i18n-tasks/commit/abd3d6ffb4e96d3cb6845b6ac466838165857fa8

PR #381, opened 2021-06-16
Allow relative keys in any Ruby object

### LINKED ISSUE #375: Allow for shallow resolution of custom relative i18n keys
Using the noticed gem, I have the following I18n use:

```ruby
class MyNotice < ApplicationNotification
  deliver_by :twilio, format: :to_twilio

  def to_twilio
    I18n.t('.super_key')
  end
end
```

I18n resolves this key to `notifications.my_notice.super_key

To use it with I18n, I simply added `/app` to the configuration

```yaml
  ## Root directories for relative keys resolution.
  relative_roots:
    - app/components
    - app/controllers
    - app/helpers
    - app/mailers
    - app/views
    - app
```

I18n-tasks thinks this is a controller action, and resolves this key to `notifications.my_notice.to_twilio.super_key` which is not what Rails / I18n itself does.

How do I tell I18n-tasks to ignore the method name here?

javierm:
When using relative keys in controllers and mailers, it usually makes sense to assume the name of the method (such as "index" or "new") should be part of the absolute key, since they generally follow the "one method == one action" pattern and there's a view file associated with this method/action.

However, in other classes, like presenters, decorators, components, or any other plain old Ruby object, that isn't usually the case. For instance, in components the view file is associated with the whole Ruby class and not just one method.

So now the calling method is only included as part of the key in classes where it's determined the key should be relative to the method.

Closes #375.

P.S. It's the first time I check the code of this project and I basically don't know what I'm doing :smile:, so any help/advice is appreciated.


lib/i18n/tasks/scanners/relative_keys.rb
```diff
@@ -42,7 +42,7 @@ def path_root(path, roots)
         def prefix(normalized_path, calling_method: nil)
           file_key       = normalized_path.gsub(%r{(\.[^/]+)*$}, '').tr(File::SEPARATOR, DOT)
           calling_method = calling_method.call if calling_method.respond_to?(:call)
-          if calling_method&.present?
+          if PatternScanner.new.key_relative_to_method?(normalized_path)
```
glebm, line 45: 1. It'd probably best to move `key_relative_to_method?` here instead of instantiating a `PatternScanner`.
2. If `calling_method` is `nil`, the result will have a trailing dot, which is probably wrong / unintuitive. I wonder what i18n actually generates in this case.

javierm:
The current approach isn't compatible with existing applications which assume the method name will be part of the generated translation key. Any advice about how to handle this scenario?

glebm:
Perhaps we could make the regexp for including the method name a setting, e.g. `relative.include_method_name_path_regex`.

For backwards compatibility, it can default to `.*`, or we can use the new default and bump major version.

glebm:
Rather than a regex settings, a list of directories would probably be better, as we already have `search.relative_roots`

javierm:
@glebm I've updated the pull request based on your comments. Again, I'm not familiar with this code, so any suggestions are much appreciated :wink:.

javierm:
I forgot to mention I'm using an `exclude` option instead of `include` option because it seemed easier to configure; using `include` in another application would probably always require adding `app/mailers` and `app/controllers` there :thinking:.

developius:
I'm interested in using this option with View Components [using the sidecar setup](https://viewcomponent.org/guide/generators.html#place-the-view-in-a-sidecar-directory), where you end up with this:

```
app/components/my_component/
│   my_component.html.erb
│   my_component.en.yml
│   my_component.js
│   my_component.scss
```

Is there a way for the `relative_exclude_method_name_paths` option to work like this?

PR #384, opened 2021-07-24
Add deepl_host and deepl_version to translation config

nm:
In order to use Deepl API Free, it is necessary to set an alternative API host. This PR adds support for Deepl host and version as optional configuration parameters.

Reference: https://github.com/wikiti/deepl-rb#usage


glebm:
Thanks!

Issue #331, opened 2019-08-13
Not ignoring pluralisation


i2chris:
It ignores datetime keys but the missing pluralisation keys still appear in the missing translations section.  Is there a specific rule needed to ignore pluralisations?

```
ignore_missing:
  - '{datetime, activerecord, date, datetime,i18n, number, devise, spree, support, time, helpers}.*'
```

```
Missing translations (13) | i18n-tasks v0.9.29
+--------+------------------------------------------------+----------------------------------+
| Locale | Key                                            | Value in other locales or source |
+--------+------------------------------------------------+----------------------------------+
|   pl   | cannot_register_samples.counting               | few, many                        |
|   pl   | datetime.distance_in_words.about_x_hours       | few, many                        |
|   pl   | datetime.distance_in_words.about_x_months      | few, many                        |
|   pl   | datetime.distance_in_words.about_x_years       | few, many                        |
|   pl   | datetime.distance_in_words.almost_x_years      | few, many                        |
|   pl   | datetime.distance_in_words.less_than_x_minutes | few, many                        |
|   pl   | datetime.distance_in_words.less_than_x_seconds | few, many                        |
|   pl   | datetime.distance_in_words.over_x_years        | few, many                        |
|   pl   | datetime.distance_in_words.x_days              | few, many                        |
|   pl   | datetime.distance_in_words.x_minutes           | few, many                        |
|   pl   | datetime.distance_in_words.x_months            | few, many                        |
|   pl   | datetime.distance_in_words.x_seconds           | few, many                        |
|   pl   | you_are_adding_items_from_the_trolley.counting | few, many                        |
+--------+------------------------------------------------+----------------------------------+
```

nitsujri:
Language specific `ignore_missing` doesn't work yet: https://github.com/glebm/i18n-tasks/issues/339

davidwessman:
@glebm Closed by #389 ?

PR #389, opened 2021-09-03
Fix not ignoring missing for pluralization

pedrozath:
References #331

When you add a pattern for ignoring missing, it is currently not ignoring the pluralization for that pattern.
This commit fixes that, by skipping plural nodes that were ignored in the configuration for `ignore_missing`.

### DIFF
--- a/lib/i18n/tasks/missing_keys.rb
+++ b/lib/i18n/tasks/missing_keys.rb
@@ -65,7 +65,7 @@ module I18n::Tasks
         plural_nodes data[locale] do |node|
           children = node.children
           present_keys = Set.new(children.map { |c| c.key.to_sym })
-          next if ignore_key?(node.key_names.first, :missing)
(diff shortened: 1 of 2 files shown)


yorsant:
References #331

lib/i18n/tasks/missing_keys.rb
```diff
@@ -65,6 +65,7 @@ def missing_plural_forest(locales, _base = base_locale)
         plural_nodes data[locale] do |node|
           children = node.children
           present_keys = Set.new(children.map { |c| c.key.to_sym })
+          next if ignore_key?(node.key_names.first, :missing)
```
glebm, line 68:
```suggestion
          next if ignore_key?(node.full_key(root: false), :missing)
```

glebm (changes requested):
Can you please also add or update the test for this?

pedrozath:
@glebm thanks for the reviewing this! just updated the test :)

PR #391, opened 2021-09-07
Use sha for interpolation replacement

### LINKED ISSUE #390: Google translate is translating `zxzxzx` from 'en' -> 'fr'
In order to avoid translating any interpolation, `i18n-tasks` replaces the interpolation with `zxzxzx` (plus a counter if there are many interpolated variables). Google translate however has a strange bug when `zxzxzx` is asked to translate from `en` to `fr`, it translates it to `zxzxxx`. So afterwards, when the interpolation is supposed to replace `zxzxzx` with the original interpolation, it fails because is is now `zxzxxx`. And so french locales don't have the correct interpolation and also have the leftover `zxzxxx` strings.

I submitted feedback to google that this is wrong (or at least this definitely _seems_ wrong to me), but there is no timeframe when that will be corrected on their end.

Screen Shot 2021-09-07 at 1 02 46 PM

npresco:
Fixes #390

Uses a digest of the interpolation variable to replace/restore during translation rather than `zxzxzx` plus a counter for the current interpolated variable. We will be sending more characters to google translate with this fix but this seemed an acceptable trade-off.

### DIFF
--- a/lib/i18n/tasks/translators/base_translator.rb
+++ b/lib/i18n/tasks/translators/base_translator.rb
@@ -94,7 +94,6 @@ module I18n::Tasks
       end
 
       INTERPOLATION_KEY_RE = /%\{[^}]+}/.freeze
-      UNTRANSLATABLE_STRING = 'zxzxzx'
 
       # @param [String] value
       # @return [String] 'hello, %{name}' => 'hello, <round-trippable string>'
--- a/lib/i18n/tasks/translators/base_translator.rb
+++ b/lib/i18n/tasks/translators/base_translator.rb
@@ -99,10 +99,8 @@ module I18n::Tasks
       # @param [String] value
       # @return [String] 'hello, %{name}' => 'hello, <round-trippable string>'
       def replace_interpolations(value)
-        i = -1
-        value.gsub INTERPOLATION_KEY_RE do
-          i += 1
-          "#{UNTRANSLATABLE_STRING}#{i}"
+        value.gsub INTERPOLATION_KEY_RE do |m|
+          Digest::SHA1.hexdigest m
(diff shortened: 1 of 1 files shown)


lib/i18n/tasks/translators/base_translator.rb
```diff
@@ -94,15 +94,12 @@ def parse_value(untranslated, each_translated)
...
 
       # @param [String] value
       # @return [String] 'hello, %{name}' => 'hello, <round-trippable string>'
       def replace_interpolations(value)
-        i = -1
-        value.gsub INTERPOLATION_KEY_RE do
-          i += 1
-          "#{UNTRANSLATABLE_STRING}#{i}"
+        value.gsub INTERPOLATION_KEY_RE do |m|
+          Digest::SHA1.hexdigest m
```
glebm, line 102: Missing a `require` for `Digets::SHA1`

glebm:
I'm not sure how robust this is: There is a chance that a substring of the SHA-1 hash is an actual word, which may confuse Google Translate.

Perhaps we can simply use a different token, something other than `zxzxzx`?

npresco:
While there may be a substring that is an actual word inside the hash, since the hash is being sent as one word, I don't believe it will attempt to translate it. Then again the entire point of this PR is because google translate is translating strange combinations of letters, or at least behaving strangely. So it's a good point.

Using a token sounds like a plan, out of curiosity, why include any alphabet character at all? `X___` vs `__`

npresco:
Closing in favor of #392

glebm:
> out of curiosity, why include any alphabet character at all?

Without a character I get this for Hebrew:

image

PR #396, opened 2021-11-08
Allow forward slash in locale key

### LINKED ISSUE #380: Not detecting keys with slashes
It looks like keys with slashes are not detected. Is there a better way to handle these?

```
t("activerecord.attributes.product/categories.name")
```

ActiveRecord uses slashes for both namespaces and nested associations.

Note: This is a duplicate of #377 but that was closed. Not certain why.

ryanb:
This adds `/` as a valid character in translation keys.

Closes #380


glebm:
Thanks!

ryanb:
Fastest PR merge I've seen. Thanks!

Issue #354, opened 2020-06-08
Unused task producing regexperror: premature end of char-class


acetinick:
Hi,

I am looking to use the unused task across my very large Rails to find unused locales.  I have a lot of dynamic locales in the code,  as assume this might be causing errors.

Any way i can work out what particularly is causing the issue with the task?

I have run this task using with and without the "--no-strict" parameter.

Below is the detailed output:
![image](https://user-images.githubusercontent.com/571649/84085521-6e88ef80-a9ee-11ea-9863-84b39180ae93.png)

Ruby 2.6.6
Rails 6.0.3.1
i18n-tasks 0.9.31

tomaadland:
I have the same problem with dynamic locales. Setting `search.strict` to `false` is not working. 

@acetinick did you find a solution for this ?

tomaadland:
After a few hours of debugging I finally found the problem in my source. 

One of the dynamic locales is an array lookup: `t("loopup.something.in.#{somearray["element"]}")` when this is compiled to key pattern it ends up with this result `(lookup\.something\.in\.somearray[|)` which makes the conversion to Regexp fail. The open bracket `[` is the culprit here. 

Solution for my problem was to remove the array lookup into a helper method.

PR #397, opened 2021-11-20
Allow hash literals in dynamic keys

otoyo:
Fix https://github.com/glebm/i18n-tasks/issues/354

Hash literals in dynamic keys such as `t("loopup.something.in.#{somearray["element"]}")` are not interpreted correctly.
This patch allows hash literals in dynamic keys.

### DIFF
--- a/lib/i18n/tasks/scanners/ruby_key_literals.rb
+++ b/lib/i18n/tasks/scanners/ruby_key_literals.rb
@@ -20,7 +20,7 @@ module I18n::Tasks::Scanners
       literal
     end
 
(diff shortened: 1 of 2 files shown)


glebm:
Thanks! :+1:

glebm:
I've migrated CI from Travis to GitHub actions and found that this actually breaks support for pattern keys with the scope argument (e.g. `t "key_in_erb", scope: "scope"`).

Can you please have a look?

https://github.com/glebm/i18n-tasks/runs/4551782658?check_suite_focus=true

otoyo:
@glebm 
I fixed this problem at https://github.com/glebm/i18n-tasks/pull/405
But CI still fail by another problem.

PR #405, opened 2021-12-17
Fix "#{double_quoted["hash"]}" pattern

otoyo:
Fix https://github.com/glebm/i18n-tasks/pull/397#issuecomment-996083497


glebm:
Thanks!

PR #414, opened 2022-01-18
Include Ruby 3.1 in github workflow test matrix

colinbruce:
Update the test matrix in the Github workflow to include Ruby 3.1

This also required updating the `ruby/setup-ruby` to `v1`


colinbruce:
This is step one of trying to diagnose the issue I'm having with aliased yaml - as it only occurred when we switched to Ruby 3.1

PR #415, opened 2022-01-18
Ensured Aliased values work

### LINKED ISSUE #412: Aliases failing in missing_keys
Since updating to Ruby 3.1 we are running into an issue with Psych not allowing aliases by default https://bugs.ruby-lang.org/issues/17866

This manifests as 
```
I18n::Tasks::CommandError:
  yaml parse error: Unknown alias: applicant_error (file: config/locales/en/activemodel.yml)
```

Changing https://github.com/glebm/i18n-tasks/blob/main/lib/i18n/tasks/data/adapter/yaml_adapter.rb#L12 to
```ruby
YAML.safe_load(str, **(options || {}), aliases: true)
```
Allows us to load and run I18n-tasks.

I'm trying to add a test but I'm struggling with creating fake data that will test this behaviour

Can anyone provide any guidance on where to be creating this?

colinbruce:
This updates the YamlAdapter to use safe_load with standard defaults, this ensures compatibility with Psych 4.0 that was made standard in Ruby 3.1, while still being compatible with older versions of ruby

[closes #412]

### DIFF
--- a/spec/file_system_data_spec.rb
+++ b/spec/file_system_data_spec.rb
@@ -67,15 +67,16 @@ RSpec.describe 'File system i18n' do
     after { TestCodebase.teardown }
 
     it '#get' do
-      data.config = { read: ['a.yml', '{b,c}.yml'] }
+      data.config = { read: ['a.yml', '{b,c}.yml', 'd.yml'] }
       TestCodebase.setup(
         'a.yml' => { en: { a: 1 } }.stringify_keys.to_yaml,
         'b.yml' => { en: { b: 1 } }.stringify_keys.to_yaml,
-        'c.yml' => { en: { c: 1 } }.stringify_keys.to_yaml
+        'c.yml' => { en: { c: 1 } }.stringify_keys.to_yaml,
+        'd.yml' => "---\n:en:\n :d: &d test\n :d_alias: *d\n"
       )
       TestCodebase.in_test_app_dir do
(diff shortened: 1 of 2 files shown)


spec/file_system_data_spec.rb
```diff
@@ -67,15 +67,16 @@
...
 
     it '#get' do
-      data.config = { read: ['a.yml', '{b,c}.yml'] }
+      data.config = { read: ['a.yml', '{b,c}.yml', 'd.yml'] }
       TestCodebase.setup(
         'a.yml' => { en: { a: 1 } }.stringify_keys.to_yaml,
         'b.yml' => { en: { b: 1 } }.stringify_keys.to_yaml,
-        'c.yml' => { en: { c: 1 } }.stringify_keys.to_yaml
+        'c.yml' => { en: { c: 1 } }.stringify_keys.to_yaml,
+        'd.yml' => "---\n:en:\n :d: &d test\n :d_alias: *d\n"
```
glebm, line 75: This could be made a bit more readable by using a heredoc:

```suggestion
        'd.yml' => <<<~YAML
          ---
          :en:
            :d: &d test
            :d_alias: *d
        YAML
```

colinbruce: Noted - I was trying to match the single-line format of the hashes!
Clarified and simplified the HEREDOC YAML to match real-life YAML structures

PR #416, opened 2022-02-02
Adds experimental ErbAstScanner

davidwessman:
- Create a basic AST for all dynamic and code calls in ERB based on the
  Temple::ERB::Parser and convert results to Parser::AST::Nodes to be
  able to reuse a lot of RubyAstScanner.
- It does not handle magic comments (maybe it would make sense to do
  them with regex instead?
- It will not get the correct position or line position, only row
  number.


glebm:
Absolutely, anything that replaces the current regexp parser is an improvement!

davidwessman:
> > Do you think something like this would be useful?
> 
> 
> 
> Absolutely, anything that replaces the current regexp parser is an improvement!

Cool 😊 it passes the test suite locally as is. But maybe it needs to be added as an opt-in.

I was also thinking if we could restructure the AST-matching to allow adding custom matchers like support for Rails model_name.human and human_attribute_name

glebm:
>  But maybe it needs to be added as an opt-in.

It can be the default (assuming comments and file positions are implemented), and we'll bump the minor version number.

> I was also thinking if we could restructure the AST-matching to allow adding custom matchers like support for Rails model_name.human and human_attribute_name

Yeah, I'm sure that'd be helpful for lots of folks.

davidwessman:
> It can be the default (assuming comments and file positions are implemented), and we'll bump the minor version number.

I will try to think if I can figure out the file positions, must look through the Shopify-gems if they have any ideas. 
`temple` removes too much information in its parsing :/


Would it be okay to only support comments that are in ERB-tags?

```erb
<!-- i18n-tasks-use t('activerecord.models.company') --> <!-- this will be ignored -->
<% # i18n-tasks-use t('activerecord.models.company') %> <!-- this will be a magic comment -->
<h1><%= Company.model_name.human %></h1>
```

glebm:
> Would it be okay to only support comments that are in ERB-tags?

Yes, this is how it is currently as well https://github.com/glebm/i18n-tasks/blob/d51327cb48870f984303da55fe1a0605cf0e70d3/lib/i18n/tasks/scanners/pattern_scanner.rb#L18

lib/i18n/tasks/scanners/ruby_ast_scanner.rb
```diff
@@ -204,6 +194,41 @@ def make_buffer(path, contents = read_file(path))
...
+
+      magic_comments.flat_map do |comment|
+        @parser.reset
+        associated_node = comment_to_node[comment]
+        @call_finder.collect_calls(
+          @parser.parse(make_buffer(path, comment.text.sub(MAGIC_COMMENT_PREFIX, '').split(/\s+(?=t)/).join('; ')))
+        ) do |send_node, _method_name|
+          # method_name is not available at this stage
+          send_node_to_key_occurrence(send_node, nil, location: associated_node || comment.location)
+        end
```
davidwessman, line 219: I think we could replace the parsing here with the `LocalRubyParser` I needed in the erb parser.

davidwessman:
@glebm I think it works now 👀

lib/i18n/tasks/scanners/ruby_ast_scanner.rb
glebm: Yep sounds good 👍

glebm (approved):

glebm:
Clean and easy-to-read code with tests, amazing contribution 👍

davidwessman:
> Clean and easy-to-read code with tests, amazing contribution 👍

Glad to hear 🙂 
Please ping me if you need anything else from me on this PR.

I will start to tinker with configurable AST-call-finders to add support for `human_attribute_name` and `model_name.human` for Rails 🙂

davidwessman:
> Clean and easy-to-read code with tests, amazing contribution 👍

Glad to hear 🙂 
Please ping me if you need anything else from me on this PR.

I will start to tinker with configurable AST-call-finders to add support for `human_attribute_name` and `model_name.human` for Rails 🙂

PR #421, opened 2022-03-22
Repair emojis after dump of yaml

Liberatys:
This pull request restores the encoded emojis right after the dump.
This allows developers to use emojis in their YAML translations which was a big issue for as at @renuo. As we would have to manually repair the emojis after normalization.

The code to restore the emojis was taken/extracted from https://github.com/ruby/psych/issues/371#issuecomment-947509865. 

This pr can be seen as a quickfix solution for the issue. We are working on a mergeable pr for psych. A solution in this repo seems more likely to be merged.

Relevant issues: https://github.com/ruby/psych/issues/371

### DIFF
--- a/lib/i18n/tasks/data/adapter/yaml_adapter.rb
+++ b/lib/i18n/tasks/data/adapter/yaml_adapter.rb
@@ -5,6 +5,8 @@ module I18n::Tasks
   module Data
     module Adapter
       module YamlAdapter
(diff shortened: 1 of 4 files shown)


glebm:
This workaround is fine by me. The actual fix should probably be in [libyaml](https://github.com/yaml/libyaml), not psych.

Liberatys:
True ^^

PR #425, opened 2022-03-24
ErbAst: Handles block calls

### LINKED ISSUE #424: With the new parser, some key usage is not found
With the update to version 1.0.0 (and still with 1.0.1), I started to get some new reported unused keys. I have confirmed that these are in fact used, and it seems they're all used in keyword arguments for method calls that take blocks. For example, in the following code the `.edit` key is not found and its corresponding translation is reported as unused.

```
<%= link_to(edit_foo_path(foo), title: t(".edit")) do %>
  <i class="fa fa-edit icon-fa"></i>
<% end %>
```

davidwessman:
- Fixes #424
- Adds more tests for erb blocks and relative keys

- Occurence: Adds keys to inspect string

### DIFF
--- a/lib/i18n/tasks/scanners/results/occurrence.rb
+++ b/lib/i18n/tasks/scanners/results/occurrence.rb
@@ -48,7 +48,7 @@ module I18n::Tasks
         # rubocop:enable Metrics/ParameterLists
 
         def inspect
-          "Occurrence(#{@path}:#{@line_num}:#{@line_pos}:#{@pos}:#{@raw_key}:#{@default_arg}:#{@line})"
+          "Occurrence(path: #{@path}, line_num: #{@line_num}, line_pos: #{@line_pos}, pos: #{@pos}, raw_key: #{@raw_key}, default_arg: #{@default_arg}, line: #{@line})"
         end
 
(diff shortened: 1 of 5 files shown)


lib/i18n/tasks/scanners/results/occurrence.rb
```diff
@@ -48,7 +48,7 @@ def initialize(path:, pos:, line_num:, line_pos:, line:, raw_key: nil, default_a
         # rubocop:enable Metrics/ParameterLists
 
         def inspect
-          "Occurrence(#{@path}:#{@line_num}:#{@line_pos}:#{@pos}:#{@raw_key}:#{@default_arg}:#{@line})"
```
davidwessman, line 51: @glebm Not sure what you think about this, can revert it 🙂

glebm: `"#{@path}:#{@line_num}"` is standard notation, let's remove the field names there but keep the rest

glebm: Going to merge this for now and release 1.0.2 anyway

PR #432, opened 2022-03-26
Handle multiline ERB comments

### LINKED ISSUE #431: Parser Error when erb comment tags span multiple lines
I noticed (initially when checking 1.0.2, but confirmed still in 1.0.4) an issue where multi-line comment tags like [the following](https://github.com/djuber/parser-error-reproduction/blob/main/app/views/layouts/application.html.erb#L11-L13)

```ruby
  <%# this is a multiline comment
  it's totally fine
  nothing should fail %>
```

are recognized correctly by ERB and render correctly (don't render at all, the entire content is commented), but a parser error is raised when [running 18n tasks](https://github.com/djuber/parser-error-reproduction/blob/main/spec/i18n_spec.rb)

```
 bundle exec rspec
#<Thread:0x00005634a007a9d8 /home/djuber/reproduction/vendor/cache/ruby/3.0.0/gems/i18n-tasks-1.0.4/lib/i18n/tasks/scanners/scanner_multiplexer.rb:33 run> terminated with exception (report_on_exception is true):
/home/djuber/reproduction/vendor/cache/ruby/3.0.0/gems/parser-3.1.1.0/lib/parser/diagnostic/engine.rb:72:in `process': Error scanning app/views/layouts/application.html.erb: unterminated string meets end of file (I18n::Tasks::CommandError)


     Failure/Error: raise Parser::SyntaxError, diagnostic
     
     I18n::Tasks::CommandError:
       Error scanning app/views/layouts/application.html.erb: unterminated string meets end of file

```

Modifying the comment tag to [include a `#` on each comment line](https://github.com/djuber/parser-error-reproduction/blob/fix-template/app/views/layouts/application.html.erb#L11-L13) resolves the parser error (tests run successfully), so we can work around this if needed, but indicates a drift between the ERB syntax and that recognized by the i18n tasks.

Originally we expected this was related to https://github.com/glebm/i18n-tasks/pull/430 (since the tag starts  `<%#` and not `<% #` that was also possibly related), but the same problem was seen after updating to 1.0.4.

glebm:
Fixes #431

### DIFF
--- a/lib/i18n/tasks/scanners/erb_ast_processor.rb
+++ b/lib/i18n/tasks/scanners/erb_ast_processor.rb
@@ -70,9 +70,16 @@ module I18n::Tasks::Scanners
         node.children[2]&.type == :code &&
         node.children[3].nil?
       code_node = node.children[2]
+
+      # Prepend # to each line to make it a valid Ruby comment.
+      code = code_node.children[0].split("\n").map do |line|
+        next line if line =~ /^\s*#/
+        "##{line}"
+      end.join("\n")
+
       node.updated(
         nil,
-        [nil, nil, code_node.updated(nil, ["##{code_node.children[0]}"]), nil]
+        [nil, nil, code_node.updated(nil, [code]), nil]
       )
     end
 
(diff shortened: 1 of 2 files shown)


davidwessman (approved):
Was just getting started on this 🙂 Looking good!

PR #433, opened 2022-03-27
Adds Rails matcher for AST-scanners

davidwessman:
- AST-matching is extracted to a class.
- Adds a class for matching Rails model translations.

### DIFF
--- a/lib/i18n/tasks/scanners/ast_matchers/rails_model_matcher.rb
+++ b/lib/i18n/tasks/scanners/ast_matchers/rails_model_matcher.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+require 'i18n/tasks/scanners/results/occurrence'
+
+module I18n::Tasks::Scanners::AstMatchers
+  class RailsModelMatcher < BaseMatcher
+    def convert_to_key_occurrences(send_node, _method_name, location: send_node.loc)
+      human_attribute_name_to_key_occurences(send_node: send_node, location: location) ||
+        model_name_human_to_key_occurences(send_node: send_node, location: location)
+    end
+
+    private
+
+    def human_attribute_name_to_key_occurences(send_node:, location:)
+      children = Array(send_node&.children)
+      receiver = children[0]
+      method_name = children[1]
+
+      return unless method_name == :human_attribute_name && receiver.type == :const
+
+      value = children[2]
+
+      model_name = underscore(receiver.to_a.last)
+      attribute = extract_string(value)
+      key = "activerecord.attributes.#{model_name}.#{attribute}"
+      [
+        key,
+        I18n::Tasks::Scanners::Results::Occurrence.from_range(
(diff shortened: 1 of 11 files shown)


glebm:
Shall we merge this before #436?

davidwessman:
> Shall we merge this before #436?

No, I think we should do the comments first. Will see if I can get some more time to work on it today.

lib/i18n/tasks/scanners/ruby_ast_scanner.rb
```diff
@@ -131,7 +132,7 @@ def setup_matchers
             message: message,
             scanner: self
           )
-        end
+        end << AstMatchers::RailsModelMatcher.new(scanner: self)
```
davidwessman, line 135: @glebm Should we check some config if the rails-matcher should be used?

lib/i18n/tasks/scanners/ast_matchers/rails_model_matcher.rb
```diff
@@ -0,0 +1,63 @@
...
+      children = Array(send_node&.children)
+      return unless children[1] == :human
+
+      base_children = Array(children[0]&.children)
+      class_node = base_children[0]
+
+      return unless class_node&.type == :const && base_children[1] == :model_name
+
+      model_name = class_node.to_a.last.downcase
+      key = "activerecord.models.#{model_name}.other"
```
davidwessman, line 53: I tried to check how pluralization is done in other places, not really sure which key to use here or if we somehow can mark multiple different keys?

glebm: Pluralization is handled elsewhere, here it should probably be "activerecord.models.#{model_name}"

lib/i18n/tasks/scanners/ruby_ast_scanner.rb
glebm: Yeah, let's think of a way to make this configurable

lib/i18n/tasks/scanners/ruby_ast_scanner.rb
```diff
@@ -132,6 +133,10 @@ def setup_matchers
             scanner: self
           )
         end
+        if Array(config[:extra_matchers]).include?('rails_model')
```
glebm, line 136: Perhaps the config should be a list of class names? This way it is extensible outside of i18n-tasks

davidwessman: Hm, yes that makes sense. Like the scanners works?

glebm: Yeah

davidwessman: Done 🙂

glebm (approved):

PR #446, opened 2022-04-10
Setup overcommit for rubocop and run rubocop -a

### LINKED ISSUE #428: Automatic formatting
Hello 🙂 

I was thinking that it would be great to add some automatic formatting, rubocop is not enforced or used to format right now.

There are a few alternatives, I think we should add a commit-hook to automatically run on staged changes and:

1. Use rubocop for formatting.
1. Use [standardrb](https://github.com/testdouble/standard).
1. Use [prettier/plugin-ruby](https://github.com/prettier/plugin-ruby).

In my opinion:

1. Need to regularly update `.rubocop.yml` when new cops are added.

2. Opinionated but more seamless to update.

3. Requires javascript, so this project might not be a good candidate.

What do you think? Then we can either format all the files at ones, or do them each time it is edited.

davidwessman:
- Run rubocop -a and fix some warnings
- Adds Overcommit to run Rubocop as pre-commit
- Fixes #428

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -1,0 +1,37 @@
+# Use this file to configure the Overcommit hooks you wish to use. This will
+# extend the default configuration defined in:
+# https://github.com/sds/overcommit/blob/master/config/default.yml
+#
+# At the topmost level of this YAML file is a key representing type of hook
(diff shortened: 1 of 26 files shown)


glebm:
Thanks!

Issue #409, opened 2021-12-22
New `--config` option doesn't appear to work.


geoffharcourt:
The `--config` option to specify an alternate location for the configuration file doesn't appear to be accepted in either 0.9.35 (where it was added) or 0.9.36.

```
> i18n-tasks --config config/i18n-tasks.deployment.yml
i18n-tasks: invalid option: --config
```

glebm:
@krtschmr Can you please have a look?

davidwessman:
@geoffharcourt Is it still applicable for the latest version?

```bash
bundle exec i18n-tasks --version    
1.0.5
bundle exec i18n-tasks --config what
i18n-tasks: Config file doesn't exist: what
```

geoffharcourt:
@davidwessman I'm not sure why, but in multiple environments we're still seeing this issue:

```
❯ bundle exec i18n-tasks --version
1.0.5
❯ bundle exec i18n-tasks --config config/i18n-tasks.deployment.yml
i18n-tasks: invalid option: --config
```

This happens if I run the gem (same version) from the system installation (not from bundler) we get the same behavior.

davidwessman:
I cannot recreate this execpt for v0.9.34 when the option did not exist.

- Could it be some gem cache?
- Can you give more information about the environment?

geoffharcourt:
Hi @davidwessman I'm still confused about what's going on. I checked and I don't have any of the pre-gem versions installed:

```
❯ gem list | grep i18n-tasks
i18n-tasks (1.0.10, 1.0.9)
```

I confirmed I'm using the version associated with my local Ruby installation (MacOS, Ruby 3.1.2):

```
❯ which i18n-tasks
/Users/geoff/.asdf/shims/i18n-tasks
```

And I'm still getting an error when trying to use the flag:

```
❯ i18n-tasks --config config/i18n-tasks.deployment.yml
i18n-tasks: invalid option: --config
```

glebm:
What does `i18n-tasks --version` say?

geoffharcourt:
Hi @glebm, sorry for not including that earlier:

```
❯ i18n-tasks --version
1.0.10
```

davidwessman:
How does your i18n-tasks.yml look?

geoffharcourt:
@davidwessman I'd hope that only the one I'm referencing matters, but here's the one that's normally in place for us (`config/i18n-tasks.yml`, at the bottom is the one we're trying to use):

```
[...]

PR #455, opened 2022-05-18
Fix 'config' option usage in CLI

shev-vadim-net:
This PR fixes the #409 issue.

As I understand the `config` option is not fully added into the gem. It's used only in the `I18n::Tasks::BaseTask.new` and nowhere else. This change just removes `config` from the arguments. Now all the CLI commands work as expected when providing a custom configuration file.

### DIFF
--- a/lib/i18n/tasks/cli.rb
+++ b/lib/i18n/tasks/cli.rb
@@ -36,11 +36,12 @@ class I18n::Tasks::CLI
   def run(argv)
     argv.each_with_index do |arg, i|
       if ['--config', '-c'].include?(arg)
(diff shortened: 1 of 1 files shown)


davidwessman:
Cool!

@geoffharcourt Can you see if this solves your problems?

geoffharcourt:
Woot! This fixed it! @shev-vadim-net thanks for the find and the fix! We use i18n-tasks in two different setups, so this lets us remove a massive hack we had been using (copying config files over the canonical location).

glebm:
Thanks!

PR #457, opened 2022-06-01
Remove circular requires, fix `Kernel#require` warning

nickcampbell18:
`I18n::Tasks::Concurrent::CachedValue` was requiring itself - this caused a warning that is avoidable.

Whilst `Siblings` does technically depend on `Node`, `Node` itself has already been required upfront by `I18n::Tasks::Data::FileSystemBase`, so removing this `require` line resolves the warning.

Collectively, these two changes resolved issues for me when loading the gem using Ruby `3.1.2p20`


glebm:
Thanks!

PR #467, opened 2022-08-05
Allow dash in locale filenames when merging data

sylvieed:
A project I'm working on uses dashes in locale filenames (i.e. `user-en.yml` instead of `user.en.yml`), which was causing `i18n-tasks` to not be able to write to the files properly (even after editing the `write` setting in `config/i18n-tasks.yml`).

This adds `-` as a valid character before the `#{locale}` part of filenames.


sylvieed:
@benwbrum @saracarl when this and #469 are merged we can switch our Gemfile back to this instead of the fork

glebm (approved):

PR #471, opened 2022-08-17
Allow better_html 2.0

tagliala:
Hi,

a new major version of [better_html](https://github.com/Shopify/better-html) has been released

Here it is the changelog: https://github.com/Shopify/better-html/releases/tag/v2.0.0

All specs pass but please check if there is a change that affects i18n-tasks


afdev82:
Any news about that? It blocks me from updating dev dependencies like erb_lint.

davidwessman:
The changelog looks good to me 🙂

Issue #417, opened 2022-02-22
DeepL translates variable names


dssjoblom:
When using the DeepL API via i18n-tasks, variable names sometimes end up translated as well. From looking at the code, there seems to be an attempt to avoid this, but it doesn't work reliably. As a rough estimate, ca. 10% of variables end up translated. It also seems to be the same variables regardless of output language, as I've tried this with English to both French and German now, with the same result.

glebm:
One option would be to treat even text translations as XML.

We'd need to HTML-escape them before sending to DeepL.

With XML translations, DeepL supports this natively.

davidwessman:
@dssjoblom Can you add any example where the variable names get translated?

dssjoblom:
@davidwessman yes, sorry for the late reply on this. You can try:

```
en:
  category: "%{model} or similar"
```
and `i18n-tasks translate-missing --backend=deepl --from=en es`.

It should yield:

```
es:
  category: "%{modelo} o similar"
```

holstvoogd:
Just ran into this myself, fairly certain always setting tag_handling: 'xml' fixes it.
I'll do more testing tomorrow and make a PR if it turns out good

SaraRavasio:
hello! is there an update here? we're also experiencing the same issue of variable names being translated 👀

davidwessman:
I tried to fix it, but not sure what the effects are on all the translations, please try it out if you are able to:
https://github.com/glebm/i18n-tasks/pull/523

PR #474, opened 2022-09-26
Send DeepL translation requests in multiple batches

schorsch:
I did get an unspecified error when sending 2000+ translations to deepl. The author of the deepl gem has relased a new version where he added an specific error when hitting a limit.

This PR also handles deepl's target locale requirement for EN and PT which need their sub-version defined e.g. EN-US.
This can be a potential 'breaking' change for users still using EN as plain target, as i added an error for this case. See https://www.deepl.com/de/docs-api/translate-text/translate-text/ with their target_lang description

### DIFF
--- a/lib/i18n/tasks/translators/deepl_translator.rb
+++ b/lib/i18n/tasks/translators/deepl_translator.rb
@@ -74,13 +74,12 @@ module I18n::Tasks::Translators
       locale.to_s.split('-', 2).first.upcase
     end
 
-    # Convert 'es-ES' to 'ES' but skip specific locales
-    # This and the above should probably be done by the deepl gem
+    # Convert 'es-ES' to 'ES' but warn about locales requiring a specific variant
     def to_deepl_target_locale(locale)
       loc, sub = locale.to_s.split('-')
       if SPECIFIC_TARGETS.include?(loc)
-        fail ::I18n::Tasks::CommandError, I18n.t('i18n_tasks.deepl_translate.errors.specific_target_missing') unless sub
-
+        # Must see how the deepl api evolves, so this could be an error in the future
+        @i18n_tasks.warn_deprecated I18n.t('i18n_tasks.deepl_translate.errors.specific_target_missing') unless sub
         locale.to_s.upcase
       else
         loc.upcase
--- a/lib/i18n/tasks/translators/deepl_translator.rb
+++ b/lib/i18n/tasks/translators/deepl_translator.rb
@@ -4,6 +4,11 @@ require 'i18n/tasks/translators/base_translator'
 
 module I18n::Tasks::Translators
   class DeeplTranslator < BaseTranslator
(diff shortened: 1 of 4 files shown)


lib/i18n/tasks/translators/deepl_translator.rb
```diff
@@ -60,11 +69,24 @@ def no_results_error_message
...
+    def to_deepl_source_locale(locale)
       locale.to_s.split('-', 2).first.upcase
     end
 
+    # Convert 'es-ES' to 'ES' but skip specific locales
+    # This and the above should probably be done by the deepl gem
+    def to_deepl_target_locale(locale)
+      loc, sub = locale.to_s.split('-')
+      if SPECIFIC_TARGETS.include?(loc)
+        fail ::I18n::Tasks::CommandError, I18n.t('i18n_tasks.deepl_translate.errors.specific_target_missing') unless sub
```
glebm, line 82: The API docs say that the the language without a region is supported for backwards compatibility. Perhaps we should issue a warning instead?

schorsch: yes that may be a bit nicer, i'll convert that to a warn_deprecated, as thats what i found in other regions of the gem.

schorsch:
Btw. i get an error using a deepl pro key for the spanish test translations:

```
     Failure/Error: expect(task.t('common.hello', 'es')).to eq(text_test[2])
     
       expected: "Hola, %{user} ¡O'Neill! Como estas?"
            got: "¡Hola, %{user} O'Neill! ¿Cómo estás?"
```

are you running those test too maybe with the free api and it has a different output? Else i would fix those tests too.

glebm:
Yeah these tests can be a bit flaky

lib/i18n/tasks/translators/deepl_translator.rb
```diff
@@ -4,6 +4,11 @@
 
 module I18n::Tasks::Translators
   class DeeplTranslator < BaseTranslator
+    # max allowed texts per request
+    BATCH_SIZE = 50
```
glebm, line 8: 50 seems a bit small, an average Rails app has 1,500+ keys. Perhaps we can increase this?

schorsch: yes, i know, but it ran with multiple request without problems for 4.5k keys. 
The limit itself is bound to the deepl api restriction: ' Up to 50 texts can be sent for translation in one request.'

glebm:
By the way, we're not currently handling placeholders correctly. This is tracked in https://github.com/glebm/i18n-tasks/issues/417

schorsch:
Thanks for merging!

> By the way, we're not currently handling placeholders correctly. This is tracked in #417

Thanks for the hint, i'll keep an ey on that, so far i had not had any exceptions my target app. Hard to tell how deepl maybe fixed any issues.

PR #483, opened 2022-12-13
Minor tweaks to the template

blooper05:
Hi! Thank you for a cool gem 😆 

I made a few tweaks to the templates to make comment-ining easier and the code style a little more consistent 🎨 

Please check my changes and merge them if you like them. Thanks!

### DIFF
--- a/templates/config/i18n-tasks.yml
+++ b/templates/config/i18n-tasks.yml
@@ -94,7 +94,7 @@ search:
   ##     User.human_attribute_name(:email) and User.model_name.human
   ##
   ## To implement your own, please see `I18n::Tasks::Scanners::AstMatchers::BaseMatcher`.
-  <%# I18n::Tasks.add_ast_matcher('I18n::Tasks::Scanners::AstMatchers::RailsModelMatcher') %>
+  # <%# I18n::Tasks.add_ast_matcher('I18n::Tasks::Scanners::AstMatchers::RailsModelMatcher') %>
 
   ## Multiple scanners can be used. Their results are merged.
   ## The options specified above are passed down to each scanner. Per-scanner options can be specified as well.
--- a/templates/config/i18n-tasks.yml
+++ b/templates/config/i18n-tasks.yml
@@ -1,5 +1,6 @@
 # i18n-tasks finds and manages missing and unused translations: https://github.com/glebm/i18n-tasks
 
+---
 # The "main" locale.
 base_locale: en
 ## All available locales are inferred from the data by default. Alternatively, specify them explicitly:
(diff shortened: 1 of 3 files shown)


templates/config/i18n-tasks.yml
```diff
@@ -94,7 +95,7 @@ search:
   ##     User.human_attribute_name(:email) and User.model_name.human
   ##
   ## To implement your own, please see `I18n::Tasks::Scanners::AstMatchers::BaseMatcher`.
-  <%# I18n::Tasks.add_ast_matcher('I18n::Tasks::Scanners::AstMatchers::RailsModelMatcher') %>
+  # <%# I18n::Tasks.add_ast_matcher('I18n::Tasks::Scanners::AstMatchers::RailsModelMatcher') %>
```
glebm, line 98: I'm not sure how this makes commenting out easier?

templates/config/i18n-tasks.yml
```diff
@@ -115,16 +116,16 @@ search:
 #     formality: prefer_less
 ## Do not consider these keys missing:
 # ignore_missing:
-# - 'errors.messages.{accepted,blank,invalid,too_short,too_long}'
-# - '{devise,simple_form}.*'
+#   - 'errors.messages.{accepted,blank,invalid,too_short,too_long}'
```
glebm, line 119: Similarly here, this doesn't make commenting out easier. The extra identation is not needed in YAML.

templates/config/i18n-tasks.yml
blooper05: 1. First of all, I think it is easier to visually understand that *this block* is commented out if it is also commented out as YAML like this.
2. The other parts are also in the format of eRuby-style comments + YAML-style comments ([here](https://github.com/glebm/i18n-tasks/blob/deba8884a20b05e4df65ca99ab3a5f4887815a47/templates/config/i18n-tasks.yml#L34-L35) or [here](https://github.com/glebm/i18n-tasks/blob/deba8884a20b05e4df65ca99ab3a5f4887815a47/templates/config/i18n-tasks.yml#L142-L153) etc), so there is also consistency.
3. I think most people use the comment in/out feature or plugin in their favorite editors. In that case, I think it's possible (and I've got it to happen) that when you try to comment this block in on a block-by-block basis, it gets commented out the other way around.

templates/config/i18n-tasks.yml
blooper05: This part is where I've made fixes related to the consistency of code style rather than ease of commenting out/in.

Yes, you're right, this indentation is not required.
But as we can see [here](https://github.com/glebm/i18n-tasks/blob/deba8884a20b05e4df65ca99ab3a5f4887815a47/templates/config/i18n-tasks.yml#L17-L21) or [here](https://github.com/glebm/i18n-tasks/blob/deba8884a20b05e4df65ca99ab3a5f4887815a47/templates/config/i18n-tasks.yml#L77-L81) etc, array lists are basically written with the indentation in this YAML file, so I just thought it would be better to make it consistent with them as well.

PR #487, opened 2023-01-23
Ignore non-html interpolations in DeepL Translator

PolinaStamenova:
As described in [DeepL xml docs](https://www.deepl.com/docs-api/xml) `ignore_tags` can only be used when `tag_handling=xml`

It worked in the Spanish example, because it is not translating user in this context.

Screenshot 2023-01-23 at 18 34 35

We discovered this bug by translating a bigger project and it happened to more than the half of our translations.
As you can see we added regression test. Enabling `xml` does  work in our code base without any side effects.

### DIFF
--- a/lib/i18n/tasks/translators/deepl_translator.rb
+++ b/lib/i18n/tasks/translators/deepl_translator.rb
@@ -37,7 +37,7 @@ module I18n::Tasks::Translators
     def options_for_translate_values(**options)
       extra_options = @i18n_tasks.translation_config[:deepl_options]&.symbolize_keys || {}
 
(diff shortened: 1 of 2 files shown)


glebm:
I think this implies that the input to DeepL should be valid XML (or perhaps HTML), so you should also HTML-escape non-HTML values when sending them to DeepL and then HTML-unescape the results.

davidwessman:
I just saw this PR after making my own :/ 
@glebm do you have examples of any values that we need to escape before sending to DeepL?

davidwessman:
This is related to #417

davidwessman:
@PolinaStamenova can you try it out on main now?

glebm:
Superceded by #523

PR #495, opened 2023-04-16
chore: add Dependabot for Actions and Gems

nschonni:
Noticed a few of the Actions were behind, so thought Dependabot would be better than a few on-off PRs


davidwessman (approved):
I really like working with dependabot 👍🏻

PR #496, opened 2023-04-16
ci: add Ruby 3.2 to test matrix

nschonni:
Wasn't sure about the minor version pinning, but I kept the same style as the others

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: ['3.2.0', '3.1.0', '3.0.1', '2.7.3', '2.6.7']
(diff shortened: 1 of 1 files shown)


davidwessman (approved):
Maybe it would be better to have them without patch-versions

.github/workflows/tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby-version: ['3.1.0', '3.0.1', '2.7.3', '2.6.7']
+        ruby-version: ['3.2.0', '3.1.0', '3.0.1', '2.7.3', '2.6.7']
```
nschonni, line 29: Alternate if you want to go to MAJOR.MINOR only
```suggestion
        ruby-version: ['3.2', '3.1', '3.0', '2.7', '2.6']
```

PR #509, opened 2023-05-31
Support partially dynamic segments

### LINKED ISSUE #508: Partially-dynamic keys are reported as unused
Hi, thank you for the work on this awesome tool!

Keys like "a.b.c-#{dynamic_portion}.d" are reported as unused, rather than the dynamic portion being ignored/marked as used. I am using `--no-strict` with the `unused` command. Is there a way to make these dynamic keys work?

brandoncc:
### What it is

Adds matching for dynamic portions of a segment. This is in addition to the current dynamic matching which matches only entire segments.

### Why

Currently, "cats.tenderlove-bio.name" would be marked unused even if the source code contains  `t "cats.#{cat}-bio.name"`.

Closes #508

### How

The library uses two special characters in regex matches right now: `*` and `:`. I combined these characters to create a "lazy match" similar the the way you combine `.+` and `?` to lazily match in standard regular expressions.

I created tests which verify that the partial match is only done in a single segment, rather than matching across segments.

Since there isn't currently a way to change which replacement string is used from `":"`, I changed the default value to the new `"*:"`.

### Questions

- What are the repercussions of changing the default replacement value?
  - In particular, I don't know if this has any negative repercussions on [this line](https://github.com/brandoncc/i18n-tasks/blob/da7d6df96cc5510cd52aa8134995bc2161950551/lib/i18n/tasks/used_keys.rb#L148) and other lines that use this variable.
- Is this likely to have any negative effects on existing users of the library?
  - I think it might be a good idea to introduce an "opt-in" option in the configuration so that existing users don't see a change in behavior. This option could become the default in a future version. Alternatively, this PR could be a breaking change and therefore warrant a major version bump according to semver.

### DIFF
--- a/lib/i18n/tasks/key_pattern_matching.rb
+++ b/lib/i18n/tasks/key_pattern_matching.rb
@@ -21,6 +21,7 @@ module I18n::Tasks::KeyPatternMatching
   # In patterns:
   #      *     is like .* in regexs
   #      :     matches a single key
+  #      *:    matches part of a single key, similar to a lazy regex
   #   { a, b.c } match any in set, can use : and *, match is captured
   def compile_key_pattern(key_pattern)
     return key_pattern if key_pattern.is_a?(Regexp)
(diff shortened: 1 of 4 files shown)


lib/i18n/tasks/key_pattern_matching.rb
```diff
@@ -21,6 +21,7 @@ def compile_patterns_re(key_patterns)
   # In patterns:
   #      *     is like .* in regexs
   #      :     matches a single key
+  #      *:    matches part of a single key, similar to a lazy regex
```
glebm, line 24:
```suggestion
  #      *:    matches part of a single key, equivalent to `[^.]+?` regex
```

glebm (approved):

glebm:
Thanks!

PR #512, opened 2023-07-11
Enable use of rails model matcher in i18n-tasks.yml ast_matchers

brad-ellert:
```
# i18n-tasks.yml
  ast_matchers:
    - 'I18n::Tasks::Scanners::AstMatchers::RailsModelMatcher'
```
As of now, the above does not work, because the proper file is not being included.


brad-ellert:
Or is this something that should go in each project's local `i18n-tasks.yml`? It also seems to work if I add
```
<% require 'i18n/tasks/scanners/ast_matchers/rails_model_matcher' %>
```
à la
```
<% I18n::Tasks.add_ast_matcher('I18n::Tasks::Scanners::AstMatchers::RailsModelMatcher') %>
```
from the template. If this is the solution, then it should also be in the template. It seems odd to use ERB syntax in here as is suggested in the template rather than the YAML syntax I showed above.

davidwessman:
I think @glebm will have to decide if it should work without the ERB-syntax 🙂

glebm:
Since it's a tiny file that does not actually depend on Rails, I think it's fine to require it by default.

Issue #348, opened 2020-04-17
private method `warn_deprecated' called


espen:
`NoMethodError: private method warn_deprecated called for #<I18n::Tasks::BaseTask>`

https://github.com/glebm/i18n-tasks/blob/master/lib/i18n/tasks/translators/google_translator.rb#L57

toxaq:
For future Googlers: You probably need to rename `api_key` to `google_translate_api_key`

PR #513, opened 2023-07-12
Fix private method called for warn deprecated

BrankoDordevic:
I have encountered a problem when trying to translate from Portuguese ("pt") to English without using the Portuguese sub locale ("pt-pt"). Which caused a `private method 'warn_deprecated' called for I18n::Tasks::BaseTask`

The use of `module_function` in `i18n/tasks/logging.rb` causes the methods to be made private and therefore makes it unable to call the `warn_deprecated` method in for example `/tasks/translators/deepl_translator.rb` and  `/tasks/translators/google_translator.rb`.

This change could potentially fix the issue in https://github.com/glebm/i18n-tasks/issues/348 as well.

### DIFF
--- a/lib/i18n/tasks/translators/base_translator.rb
+++ b/lib/i18n/tasks/translators/base_translator.rb
@@ -3,6 +3,8 @@
 module I18n::Tasks
   module Translators
     class BaseTranslator
+      include ::I18n::Tasks::Logging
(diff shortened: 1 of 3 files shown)


glebm:
Thanks! Can you please fix the lint error?

PR #519, opened 2023-08-28
feat: add OpenAI translator

jmeiss:
Hi,

Thanks a lot for that amazing gem, we love it!

We are using OpenAI to translate our application because it's doing a very good job of translating in Arabic.

So I created a translator for OpenAI/ChatGPT.

I hope this can help some people.

### DIFF
--- a/lib/i18n/tasks/translators/openai_translator.rb
+++ b/lib/i18n/tasks/translators/openai_translator.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+
+require 'i18n/tasks/translators/base_translator'
+
+module I18n::Tasks::Translators
+  class OpenAiTranslator < BaseTranslator
+    # max allowed texts per request
+    BATCH_SIZE = 50
+
+    def initialize(*)
+      begin
+        require 'openai'
+      rescue LoadError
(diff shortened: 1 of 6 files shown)


glebm (approved):

glebm:
Please address rubocop errors

lib/i18n/tasks/translators/openai_translator.rb
```diff
@@ -63,6 +63,8 @@ def translate_values(list, from:, to:, **options)
     end
 
     def translate(values, from, to)
+      # Prompt from
+      # https://github.com/ObservedObserver/chatgpt-i18n/blob/main/src/services/translate.ts
```
glebm, line 67: That repo has AGPL license, i18n-tasks has MIT license.
Please try to come up with your own prompt.

glebm:
There are still some rubocop errors. For the `Metrics/MethodLength`, feel free to disable it for that method.

lib/i18n/tasks/translators/openai_translator.rb
jmeiss: I simply removed the comment because I had already come up with a different version of the prompt. The comment was here to tell where I got the inspiration from.

PR #523, opened 2023-09-16
DeepL: Default to tag_handling: xml

### LINKED ISSUE #417: DeepL translates variable names

davidwessman:
- Also asks DeepL to ignore the <i18n> tag which is used to wrap
  interpolated variables.
- Fixes #417

### DIFF
--- a/lib/i18n/tasks/translators/deepl_translator.rb
+++ b/lib/i18n/tasks/translators/deepl_translator.rb
@@ -41,11 +41,11 @@ module I18n::Tasks::Translators
     end
 
     def options_for_html
-      { tag_handling: 'xml', ignore_tags: %w[i18n] }
+      { tag_handling: 'xml' }
     end
 
     def options_for_plain
-      { preserve_formatting: true, tag_handling: 'xml', ignore_tags: %w[i18n] }
+      { preserve_formatting: true, tag_handling: 'xml', html_escape: true }
     end
 
     # @param [String] value
--- a/lib/i18n/tasks/translators/deepl_translator.rb
+++ b/lib/i18n/tasks/translators/deepl_translator.rb
@@ -41,11 +41,11 @@ module I18n::Tasks::Translators
     end
 
     def options_for_html
-      { tag_handling: 'xml' }
+      { tag_handling: 'xml', ignore_tags: %w[i18n] }
     end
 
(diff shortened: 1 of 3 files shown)


spec/deepl_translate_spec.rb
```diff
@@ -6,15 +6,16 @@
 
 RSpec.describe 'DeepL Translation' do
   nil_value_test = ['nil-value-key', nil, nil]
-  text_test      = ['key', "Hello, %{user} O'Neill! How are you?", "¡Hola, %{user} O'Neill! ¿Cómo estás?"]
+  text_test      = ['key', "Hello, %{user} O'Neill! How are you?", "¡Hola, %{user} O'Neill! ¿Qué tal estás?"]
```
davidwessman, line 9: Not sure why this changes :/ and Spanish is not my forté, I googled it and it seems like it might be right.

glebm: Well that's definitely not right but that's OK it's just a test.

lib/i18n/tasks/translators/deepl_translator.rb
```diff
@@ -41,11 +41,11 @@ def options_for_translate_values(**options)
     end
 
     def options_for_html
-      { tag_handling: 'xml' }
+      { tag_handling: 'xml', ignore_tags: %w[i18n] }
     end
 
     def options_for_plain
-      { preserve_formatting: true }
+      { preserve_formatting: true, tag_handling: 'xml', ignore_tags: %w[i18n] }
```
glebm, line 48: Do we now need to html-escape/unescape plain text?

davidwessman: I added a suggestion on escaping + unescaping now 🙂

spec/deepl_translate_spec.rb
davidwessman: Yeah, I guess it is the users problem when using interpolation :/

PR #527, opened 2023-10-16
Set the temperature of OpenAI requests to 0 to ensure consistent translation results

czj:
Temperatures higher than 0 make OpenAI's GPT-3.5 Turbo "hallucinate" more instead of using the most suitable next word.

Setting it to 0.7 makes the translations variable and sub-optimal.


czj:
Thanks a lot !

glebm:
Thank you!

Issue #456, opened 2022-05-26
Recursively translating yaml's with lists does not work  


rscherf:
What an incredibly useful gem! Thank you for building/maintaining it! 👏

I have a somewhat complex translation yaml schema, which includes lists/sequences. Consider the following simple example:

```yaml
en:
  test:
    one: Will this translate?
    two:
    - title: How about this?
      body: Or this?
    three:
      four:
      - title: How about this level?
        body: And this?
    five:
      six: This translates
      seven:
        eight: And this does as well
```

Then, when running `i18n-tasks translate-missing`, I receive this (alphabetized for simplicity):

```yaml
de: 
  test:
    one: Wird das übersetzt?
    two:
    - title: How about this?
      body: Or this?
    three:
      four:
      - title: How about this level?
        body: And this?
    five:
      seven:
        eight: Und dies auch
      six: Dies übersetzt
```

Is there a mechanism for auto-translating all these keys, including those with a `-` dash prefix? Am I missing something? Can I force the gem to recognize this structure?

Thanks for all you do!

rscherf:
FWIW, I also tried with various `[]` and `{}`:
```
en:
  test:
    one: Will this translate?
    two: {
      title: How about this?,
      body: Or this?
    }
    three:
      four: {
        title: How about this level?,
        body: And this?
      }
    five:
      six: This translates
      seven:
        eight: And this does as well
        nine: [
          {
            title: The title,
            body: A body,
          },
          {
            title: Another title,
            body: Another body,
          },
        ]
```

to

```
de:
  test:
    five:
      seven:
        eight: Und dies auch
        nine:
        - title: The title
          body: A body
        - title: Another title
          body: Another body
      six: Dies übersetzt
    one: Wird das übersetzt?
    three:
      four:
        body: Und das?
        title: Wie wäre es mit diesem Niveau?
    two:
      body: Oder dieses?
      title: Wie wäre es damit?
```

rscherf:
And after even further testing, there the following does work:

```yaml
en:
  test:
    - Translate this
    - Then this
    - And after, this
```

Properly translates to:

```yaml
de:
  test:
    - Übersetze das
    - Dann das
    - Und danach das
```

glebm:
The translator only handles arrays and strings at the moment

If you'd like to attempt a fix by adding Hash support, the code is here

[...]

PR #531, opened 2023-11-13
Parse array of hashs for translators

georf:
With reference to ticket #456, this change now allows you to translate lists of hashes with DeepL and other translators.


glebm:
Thanks!

PR #532, opened 2023-11-20
Add a configuration for OpenAI model to use

michaelbaudino:
OpenAI provides [many models][0] (e.g. GPT-3.5 Turbo, GPT-4, GPT-4 Turbo, …), with different prices, context sizes, performance, …

This commit allows users to configure the model they want to use via the `translations.openai_model` entry in `i18n-tasks.yml`, but keeps the one that was previously hardcoded as default (`gpt-3.5-turbo`).

[0]: https://platform.openai.com/docs/models

### DIFF
--- a/lib/i18n/tasks/translators/openai_translator.rb
+++ b/lib/i18n/tasks/translators/openai_translator.rb
@@ -50,6 +50,10 @@ module I18n::Tasks::Translators
       end
     end
 
+    def model
+      @model ||= @i18n_tasks.translation_config[:openai_model].presence || "gpt-3.5-turbo"
+    end
+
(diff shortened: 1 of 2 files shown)


michaelbaudino:
I also changed the default prompt to instruct OpenAI not to translate HTML markups (`<strong>`) nor Ruby I18n variables (`%{count}`).

Would you be interested in a PR that:
:one: → changes the default system prompt with the above suggestions?
:two: → makes the system prompt configurable from `i18n-tasks.yml`?
:three: → both :one: and :two:?
:four: → none of the above?

glebm:
Sure, 3 sounds good

PR #533, opened 2023-11-20
Add configuration for OpenAI prompt

michaelbaudino:
The default prompt now includes instructions to **never** translate HTML tags (e.g. `<strong>`) nor Ruby I18n variables (e.g. `%{count}`).

This PR also makes the system prompt passed to OpenAI configurable directly from `i18n-tasks.yml`.

Incidentally, it also fixes a Rubocop offense.

This is a followup of #532.

### DIFF
--- a/lib/i18n/tasks/translators/openai_translator.rb
+++ b/lib/i18n/tasks/translators/openai_translator.rb
@@ -51,7 +51,7 @@ module I18n::Tasks::Translators
     end
 
     def model
-      @model ||= @i18n_tasks.translation_config[:openai_model].presence || "gpt-3.5-turbo"
+      @model ||= @i18n_tasks.translation_config[:openai_model].presence || 'gpt-3.5-turbo'
     end
 
     def translate_values(list, from:, to:)
--- a/lib/i18n/tasks/translators/openai_translator.rb
+++ b/lib/i18n/tasks/translators/openai_translator.rb
@@ -1,11 +1,24 @@
 # frozen_string_literal: true
 
 require 'i18n/tasks/translators/base_translator'
+require 'active_support/core_ext/string/filters'
 
 module I18n::Tasks::Translators
   class OpenAiTranslator < BaseTranslator
     # max allowed texts per request
     BATCH_SIZE = 50
+    DEFAULT_SYSTEM_PROMPT = <<~PROMPT.squish
+      You are a professional translator that translates content from the %{from} locale
+      to the %{to} locale in an i18n locale array.
+
+      The array has a structured format and contains multiple strings. Your task is to translate
+      each of these strings and create a new array with the translated strings.
+
+      HTML markups (enclosed in < and > characters) must not be changed under any circumstance.
+      Variables (starting with %%{ and ending with }) must not be changed under any circumstance.
+
+      Keep in mind the context of all the strings for a more accurate translation.
+    PROMPT
 
     def initialize(*)
       begin
@@ -54,6 +67,10 @@ module I18n::Tasks::Translators
       @model ||= @i18n_tasks.translation_config[:openai_model].presence || 'gpt-3.5-turbo'
     end
 
+    def system_prompt
+      @system_prompt ||= @i18n_tasks.translation_config[:openai_system_prompt].presence || DEFAULT_SYSTEM_PROMPT
+    end
+
     def translate_values(list, from:, to:)
       results = []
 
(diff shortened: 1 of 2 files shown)


lib/i18n/tasks/translators/openai_translator.rb
```diff
@@ -66,14 +83,11 @@ def translate_values(list, from:, to:)
...
-    def translate(values, from, to) # rubocop:disable Metrics/MethodLength
+    def translate(values, from, to)
       messages = [
         {
           role: 'system',
-          content: "You are a helpful assistant that translates content from the #{from} to #{to} locale in an i18n
-          locale array. The array has a structured format and contains multiple strings. Your task is to translate
-          each of these strings and create a new array with the translated strings. Keep in mind the context of all
-          the strings for a more accurate translation.\n"
+          content: system_prompt.gsub('{{from}}', from).gsub('{{to}}', to)
```
michaelbaudino, line 90: I would have loved using `Kernel#format` rather than `gsub` here, but the presence of `%{ and ending with }` in the default prompt prevents it: `format` considers it an interpolation named ` and ending with ` and raises an error because… well, we don't provide a value for such a variable. :sweat_smile:

glebm: You can use `%%` to escape a percent in a format string

https://github.com/ruby/ruby/blob/46ef74f270d022068def440ab6208e1648030847/sprintf.c#L436-L440

michaelbaudino: Nice, thanks: I just changed it :ok_hand:

templates/config/i18n-tasks.yml
```diff
@@ -116,6 +116,17 @@ search:
 #   # OpenAI
 #   openai_api_key: "sk-XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
 #   # openai_model: "gpt-3.5-turbo" # see https://platform.openai.com/docs/models
+#   # openai_system_prompt: # may contain `%{from}` and `%{to}`, which will be replaced by source and target locale codes, respectively, using `Kernel.format`
```
glebm, line 119: The settings in this file are formatted sothat uncommenting turns it into a valid option
Please move the comment above the line and add the appropriate multiline string sigil.

michaelbaudino: I just changed it as I think you requested :+1:

I tried to make it consistent with what was done for Deepl configuration, for both mandatory configuration lines, optional (commented) configuration lines and plain old comments (with plain old English).

I used a [folded style](https://yaml.org/spec/1.2.2/#813-folded-style) indicator (`>`) with a  "stripping" [block chomping](https://yaml.org/spec/1.2.2/#8112-block-chomping-indicator) indicator (`-`) to explicitly indicate the multiline field and how it will be processed (resulting in a `>-` marker): it seems to me it was the more logical processing method to apply, but I'm happy to change it if you prefer another one.

I hope that's what you expected.

------------------

On a side note, though, I'm pretty sure the previous implementation was still valid YAML when uncommented (despite the end-of-line comment and the lack of explicit multiline indicator :shrug:):

```ruby
YAML.safe_load(<<~YAML)
  openai_system_prompt: # may contain `%{from}` and `%{to}`, which will be replaced by source and target locale codes, respectively (using `Kernel.format`)
    You are a professional translator that translates content from the %{from} locale
    to the %{to} locale in an i18n locale array.
  
    The array has a structured format and contains multiple strings. Your task is to translate
    each of these strings and create a new array with the translated strings.
  
    HTML markups (enclosed in < and > characters) must not be changed under any circumstance.
    Variables (starting with %%{ and ending with }) must not be changed under any circumstance.
  
    Keep in mind the context of all the strings for a more accurate translation.
YAML
# => {"openai_system_prompt"=>
#      "You are a professional translator that translates content from the %{from} locale to the %{to} locale in an i18n locale array.\nThe array has a structured format and contains multiple strings. Your task is to translate each of these strings and create a new array with the translated strings.\nHTML markups (enclosed in < and > characters) must not be changed under any circumstance. Variables (starting with %%{ and ending with }) must not be changed under any circumstance.\nKeep in mind the context of all the strings for a more accurate translation."}
```

YAML is such a pain :trollface:

glebm: Indeed, YAML is gnarly. Thanks!

michaelbaudino: Thanks to you for all the work on this gem :pray: :purple_heart:

PR #536, opened 2023-11-29
Fix DeepL glossaries

michaelbaudino:
The `glossary_id` parameter was previously ignored because the `source_lang` parameter is mandatory when the `glossary_id` parameter is used.

This commit fixes this by always passing the `source_lang` (and the `target_lang`, fwiw) to DeepL `translate` queries.

### DIFF
--- a/lib/i18n/tasks/translators/deepl_translator.rb
+++ b/lib/i18n/tasks/translators/deepl_translator.rb
@@ -28,7 +28,7 @@ module I18n::Tasks::Translators
           parts,
           to_deepl_source_locale(from),
           to_deepl_target_locale(to),
-          options_with_glossary(options, from, to)
+          options.merge(language_pair_specific_options(from, to))
         )
         if res.is_a?(DeepL::Resources::Text)
           results << res.text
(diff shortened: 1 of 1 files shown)


michaelbaudino:
Those parameters should actually already been passed by `deepl-rb` :thinking:

This fix does not fix anything. Let me dig deeper to understand why my glossary seems ignored :detective:

michaelbaudino:
OK, my glossary was actually ignored because I also specified a `context` parameter to the DeepL API :confused:

According to [DeepL API documentation](https://www.deepl.com/docs-api/translate-text/translate-text/):
> The `context` parameter is an **alpha feature** […] not intended to be used in production

So my guess, is that it may be incompatible with glossaries (or at least override them / take precedence over them).

This is not something we can do anything about (I can reproduce the issue with cURL :shrug:), so what do you think we should do @glebm ?
:one: Add a warning in the README?
:two: Add a warning in the [default configuration file](https://github.com/glebm/i18n-tasks/blob/main/templates/config/i18n-tasks.yml)?
:three: Output a warning when both `context` and `glossary_id` are passed?
:four: Nothing? :sunglasses: 
:five: Anything else?

glebm:
How about 1 + 2 + 3?

PR #539, opened 2023-12-10
Use ast_matchers for default_i18n_subject matcher in config template

Bilka2:
Same change as #512 did for the rails_model_matcher. And update the config template as well.


davidwessman (approved):
Cool 😊 does the old way of requiring still work?

Bilka2:
> does the old way of requiring still work?

Yes, old configurations with the require and `add_ast_matcher` call directly in the yml keep working. It was the same for #512 with the rails model matcher.

PR #541, opened 2023-12-14
Add an isolating router

michaelbaudino:
This new router considers each source YAML file independant from each other, thus allowing each of them to include similar keys.

Here is an overview of how it works:
* when reading YAML files (as defined by patterns in the `data.read` configuration, like for other routers), it does not load them all in a common global space, but rather creates a top-level node for YAML file and then loads its content inside this node.
* when routing a given forest, it actually iterates on all top-level keys (which are source file paths) then yields the given block for each top-level key, passing also its subtree.

Note that the top-level key (which is the source file path) had to be enclosed in a character in order to prevent considering its dots (`.`) as a nesting level when de-flattening them. A mechanism to do this already existed for `{}`, `[]` and `()` but all of those enclosers have a meaning in regular expressions. Thus, this commit extends this mechanism to support `<>` and uses them as encloser for the top-level key. It also has the advantage of making this special part of the flattened key easily distinguishable by a human eye.

For example, the following YAML source files:

* `app/components/movies_component.en.yml`:

   ```yaml
   en:
     title: Movies
   ```

* `app/components/games_component.en.yml`

   ```yaml
   en:
     title: Games
   ```

… result in the following forest once loaded in memory:
```yaml
en:
  <app/components/movies_component.en.yml>:
    title: Movies
  <app/components/games_component.en.yml>:
    title: Games
```

… which in turn results in the following flattened keys:
```
en.<app/components/movies_component.en.yml>.title: Movies
en.<app/components/games_component.en.yml>.title: Games
```

… which can be sent to any translator and should come back as:
```
fr.<app/components/movies_component.en.yml>.title: Flims
fr.<app/components/games_component.en.yml>.title: Jeux
```

… finally, this router allows re-assigning each key to the appropriate target file, which gives:

* `app/components/movies_component.fr.yml`:

   ```yaml
   fr:
     title: Flims
   ```

* `app/components/games_component.fr.yml`

   ```yaml
   fr:
     title: Jeux
   ```

This implicitely adds support for ViewComponent (which suggests organizing YAML files in "sidecars" and which adds an implicit scope to keys in those files, based on their file path).

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,6 +338,33 @@ data:
 If you want to have i18n-tasks reorganize your existing keys using `data.write`, either set the router to
 `pattern_router` as above, or run `i18n-tasks normalize -p` (forcing the use of the pattern router for that run).
 
+##### Isolating router
+
+Isolating router assumes each YAML file is independent and can contain similar keys.
+
+As a result, the translations are written to an alternate target file for each source file
+(only the `%{locale}` part is changed to match target locale). Thus, it is not necessary to
+specify any `write` configuration (in fact, it would be completely ignored).
+
+This can be useful for example when using [ViewComponent sidecars](https://viewcomponent.org/guide/translations.html)
+(ViewComponent assigns an implicit scope to each sidecar YAML file but `i18n-tasks` is not aware of
+that logic, resulting in collisions):
+
+* `app/components/movies_component.en.yml`:
+   ```yaml
+   en:
+     title: Movies
+   ```
+
+* `app/components/games_component.en.yml`
+   ```yaml
+   en:
+     title: Games
+   ```
+
+This router has a limitation, though: it does not support detecting missing keys from code usage
+(since it is not aware of the implicit scope logic).
+
 ##### Key pattern syntax
 
 A special syntax similar to file glob patterns is used throughout i18n-tasks to match translation keys:
--- a/lib/i18n/tasks/data/file_system_base.rb
+++ b/lib/i18n/tasks/data/file_system_base.rb
@@ -3,6 +3,7 @@
 require 'i18n/tasks/data/tree/node'
 require 'i18n/tasks/data/router/pattern_router'
 require 'i18n/tasks/data/router/conservative_router'
+require 'i18n/tasks/data/router/isolating_router'
 require 'i18n/tasks/data/file_formats'
 require 'i18n/tasks/key_pattern_matching'
 
@@ -151,6 +152,7 @@ module I18n::Tasks
 
       ROUTER_NAME_ALIASES = {
         'conservative_router' => 'I18n::Tasks::Data::Router::ConservativeRouter',
+        'isolating_router' => 'I18n::Tasks::Data::Router::IsolatingRouter',
         'pattern_router' => 'I18n::Tasks::Data::Router::PatternRouter'
       }.freeze
       def router
@@ -170,6 +172,9 @@ module I18n::Tasks
         end.map do |path|
           [path.freeze, load_file(path) || {}]
         end.map do |path, data|
+          if router.is_a?(I18n::Tasks::Data::Router::IsolatingRouter)
+            data.transform_values! { |tree| { "<#{router.alternate_path_for(path, base_locale)}>" => tree } }
+          end
           filter_nil_keys! path, data
           Data::Tree::Siblings.from_nested_hash(data).tap do |s|
             s.leaves { |x| x.data.update(path: path, locale: locale) }
--- a/lib/i18n/tasks/data/router/isolating_router.rb
+++ b/lib/i18n/tasks/data/router/isolating_router.rb
@@ -1,0 +1,146 @@
+# frozen_string_literal: true
+
+require 'i18n/tasks/key_pattern_matching'
+require 'i18n/tasks/data/tree/node'
+
+module I18n::Tasks
+  module Data::Router
+    # Route based on source file path
+    class IsolatingRouter
+      include ::I18n::Tasks::KeyPatternMatching
+
+      attr_reader :config_read_patterns, :base_locale
+
+      def initialize(_adapter, data_config)
+        @base_locale = data_config[:base_locale]
+        @config_read_patterns = Array.wrap(data_config[:read])
+      end
+
+      # Route keys to destinations
+      # @param forest [I18n::Tasks::Data::Tree::Siblings] forest roots are locales.
+      # @yieldparam [String] dest_path
+      # @yieldparam [I18n::Tasks::Data::Tree::Siblings] tree_slice
+      # @return [Hash] mapping of destination => [ [key, value], ... ]
+      def route(locale, forest, &block)
+        return to_enum(:route, locale, forest) unless block
+
+        locale = locale.to_s
+        out = {}
+
+        forest.keys do |key_namespaced_with_source_path, _node|
+          source_path, key = key_namespaced_with_source_path.match(/\A<([^>]*)>\.(.*)/).captures
+          target_path = alternate_path_for(source_path, locale)
+          next unless source_path && key && target_path
+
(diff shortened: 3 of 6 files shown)


glebm:
I'm guessing you have a plan to address the missing keys limitation :)

michaelbaudino:
Well, to be honest: not really 😭

If I understand the problem correctly, and with my use case in mind (which is ViewComponent, but we probably want a generic-enough solution, that does not tightly couple `i18n-tasks` with ViewComponent), I see multiple problems described below (with some potential solutions).

## 1. Translations may live in multiple files

### Problem

For a given ViewComponent (let's use `MoviesComponent` as an example), its i18n translations may live in multiple files (all of them are [valid and considered part of the component sidecar](https://viewcomponent.org/guide/translations.html)):
* `app/components/movies_component/movies_component.en.yml` ([subdirectory](https://viewcomponent.org/guide/templates.html#subdirectory) method)
* `app/components/movies_component.en.yml` (locale-specific [sibling file](https://viewcomponent.org/guide/templates.html#sibling-file) method)
* `app/components/movies_component.yml` ([sibling file](https://viewcomponent.org/guide/templates.html#sibling-file) method containing multiple locales, not supported by current `isolating_router` implementation)

Thus, checking if a given translation exists (_e.g._ when encountering `t(".title")` in a component) in a tree of translations namespaced by file name is not straightforward.

### Solution 1: try all files

Look into all possible i18n files:
* `app/components/movies_component.en.yml`
* `app/components/movies_component/movies_component.en.yml`
* `app/components/movies_component.en.yaml`
* `app/components/movies_component/movies_component.en.yaml`
* `app/components/movies_component.en.json`
* `app/components/movies_component/movies_component.en.json`

Potentially also in currently unsupported files:
* `app/components/movies_component.yml`
* `app/components/movies_component/movies_component.yml`
* `app/components/movies_component.yaml`
* `app/components/movies_component/movies_component.yaml`
* `app/components/movies_component.json`
* `app/components/movies_component/movies_component.json`

It might be a bit tedious and should be configurable if we don't want to be tightly coupled to ViewComponent. For example like that:

```yaml
data:
  # regexp that defines all files sharing a common implicit i18n scope
  implicit_scope_selector: 'app/components/(%{id}/)?/%{id}(.%{locale})?.{yml,yaml,json}'
```

### Solution 2: namespace with implicit scope rather than original file path

When reading i18n files to load all translations in memory, we could namespace translations in their implicit scope (see problem 2 below to see how this implicit scope can be computed 👇) rather than in their file path (current behavior of `isolating_router`).

Thus, the following YAML file:

```yaml
# app/components/collections/movies_component.en.yml
en:
 title: Movies
```

… could be loaded in memory as the following tree:

```yaml
en:
  <collections.movies_component>:
    title: Movies
```

… rather than the current behaviour that loads the same file as follows:

```yaml
en:
  <app/components/collections/movies_component.en.yml>:
    title: Movies
```

## 1b. Routing a translation to a file must ensure consistency with source file

### Problem

If we choose solution 2 above (_i.e._ namespacing with implicit scope rather than with original file name), routing (_i.e._ finding out in which file must be written a translation based on its flat key) must now determine in which of the multiple files the translation (in target locale) should be written, since we don't know the source file it originally comes from anymore.

For example, given the following source file:

```yaml
# app/components/movies_component/movies_component.en.yml
en:
  title: Movies
```

… it would be stored in memory like that:

```yaml
en:
  <collections.movies_component>:
    title: Movies
```

The router should write translations in target locales (_e.g._ French) to `app/components/movies_component/movies_component.fr.yml` rather than `app/components/movies_component.fr.yml`, despite both files being actually valid options.

### Solution 1: namespace with both implicit scope and original file path

We could namespace in memory with **both** the source file path and the implicit scope, _e.g._ separated with a colon (`:`) or a pipe (`|`):

```yaml
en:
  <collections.movies_component:app/components/collections/movies_component/movies_component.en.yml>:
    title: Movies
```

This would allow us to parse the key:
* the "i18n scope" part would allow us properly detecting if a given translation is used (still solving problem 1 above ☝️).
* the "file path" part would allow us to know where to write a given translation.

### Solution 2: detect from which file a given flat key originates

During the routing phase, we could try to detect which file a given translation originally came from, by re-reading all source i18n files. This could lead us to choose the appropriate alternate file to write to.

I'm not sure how accurate this detection can be, but it could probably use the `implicit_scope_selector` configuration (described in solution 1 of problem 1 above ☝️):

```yaml
data:
  # regexp that defines all files sharing a common implicit i18n scope
  implicit_scope_selector: 'app/components/(%{id}/)?/%{id}(.%{locale})?.{yml,yaml,json}'
```

## 2. Implicit scope logic not known

### Problem

The implicit scope applied to each file is a logic known only by the third-party library (in my case: ViewComponent) and potentially changing in different library versions.

Thus, namespacing an i18n file content with it (as suggested in solution 2 of problem 1 above ☝️) or checking if a key exist when parsing a call to `translate` in a component or template file requires to generate this implicit scope.

### Solution

Configure implicit scope logic as a chain of regular expression substitutions applied to the file name and that would produce the implicit i18n scope to be applied to translations in this file.

For example, current [ViewComponent implicit scope](https://github.com/ViewComponent/view_component/blob/1b3747b156b8fd8a0bd058c2a4b94856add198aa/lib/view_component/translatable.rb#L19-L21) could be generated using the following configuration:

```yaml
data:
  # list of "pattern + substitution" couples passed to a chain of `gsub` calls applied to the file name
  implicit_scope_transformers:
    - ['{app/components/**/*}.{rb,en.ya?ml,erb,html,slim}', '\1.\2']
    - ['{*}_component/{*}_component.', '\1_component.']
    - ['^/', '']
    - ['/_?', '.']
```

> [!WARNING]
> This works with ViewComponent logic which involves a [chain of regular expression substitutions](https://github.com/ViewComponent/view_component/blob/1b3747b156b8fd8a0bd058c2a4b94856add198aa/lib/view_component/translatable.rb#L19-L21), but other implicit scope generation logics may not be defined using only regular expressions!
>
> For example, a logic that would force or swap character case could not be implemented using `gsub` only (since Ruby regular expressions do not support `\U` and `\L` substitution tokens, like Perl does…), neither would a logic with even more complex manipulations.

## 3. Missing keys exhaustiveness

### Problem

ViewComponent supports both translations with implicit scope (`t(".title")`) and classic translations (`t("my.global.translation")`).

In the process of checking if a translation exists for any given call to `translate` parsed, we must be able to look into both namespaced translations (as loaded when the `isolating_router` is enabled) and classic (non-namespaced) translations (as loaded when other routers are enabled).

This is currently not possible because a single `router` directive is active for each `i18n-tasks` configuration.

### Solution 1: use multiple routers in same configuration

In order to have both a namespaced (with either implicit scopes, file names or both 🤷) and non-namespaced trees loaded at the same time, we need to be able to configure different routers for different file paths, for example:

```yaml
data:
  router:
    isolating_router:
      only:
        - app/components/**/*.%{locale}.yml
    conservative_router:
      only:
        - config/locales/**/*.%{locale}.yml
```

It would probably also require the routing mechanism to detect if a given flat key is a namespace of not (_i.e._ if it is enclosed by `<>`), to eventually delegate to the appropriate router.

---------

I'd love to have other people opinions.

Am I seeing problems that do not exist?

Am I over-complicating all this?

Or would it take a lot of changes to support missing keys detection properly with `isolating_router`?

michaelbaudino:
I realize the length of my latest comment may be a bit excessive, I'm sorry :grimacing: 

But if you had an idea that is straightforward to implement, feel free to share it and I'd be happy to work on it.

glebm:
Perhaps something along the lines of how we handle mailers?

Here, we modify the key in the scanner based on the filename
https://github.com/glebm/i18n-tasks/blob/30b7703104bcfcc9a17dde6d2bb956cee554e654/lib/i18n/tasks/scanners/relative_keys.rb#L54-L65

PR #542, opened 2023-12-28
Install gem into development group only?

pglombardo:
While investigating memory usage with derailed I realized that I could save a small chunk of memory in production by moving the il8n-tasks gem to the development group.

Screenshot 2023-12-28 at 09 23 48

Should this gem be installed in the development group only?

This PR is just an update to the README Gemfile install line.


davidwessman:
I always put it in the development group 🙂

glebm:
If you use it in tests, it should also be in the test group

PR #545, opened 2024-01-16
Erb-scanner: Replaces better_html with regex

davidwessman:
- Simplify the ERB AST scanner by using the regex from erubi.
- This allows us to parse .erb that is not html as well.


glebm:
Nice!

PR #550, opened 2024-02-15
Fix key splitting failures if key contains any parenthesis char

### LINKED ISSUE #549: Bug: Missing key checks and normalization fail if keys contain single parenthesis characters
### Description
In case keys contain characters that can be used as parenthesis to escape dots, the split_key function returns erroneous results, causing tasks like normalization and 'check-missing' to fail.

### Current Situation
```ruby
> I18n::Tasks::SplitKey.split_key('a.b->c.d')
["a", "b->c.d"]
```
_Note that in the part after the '>', dots are escaped even though (1) it is a parenthesis-end character rather than a start character, (2) it's never followed by another parenthesis character._

### Expected Situation
```ruby
I18n::Tasks::SplitKey.split_key('a.b->c.d')
["a", "b->c", "d"]
```

### Solution
I'll add a PR below

tom-brouwer-bex:
Note that the complete split_key function has been refactored to be more readable, and a test has been added to check for the failures that occurred previously

Closes #549

### DIFF
--- a/lib/i18n/tasks/split_key.rb
+++ b/lib/i18n/tasks/split_key.rb
@@ -5,6 +5,11 @@ module I18n
     module SplitKey
       module_function
 
+      PARENTHESIS_PAIRS = %w({} [] () <>).freeze
+      START_KEYS = PARENTHESIS_PAIRS.to_set { |pair| pair[0] }.freeze
+      END_KEYS = PARENTHESIS_PAIRS.to_h { |pair| [pair[0], pair[1]] }.freeze
+      private_constant :PARENTHESIS_PAIRS, :START_KEYS, :END_KEYS
+
       # split a key by dots (.)
       # dots inside braces or parenthesis are not split on
       #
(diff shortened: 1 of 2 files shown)


lib/i18n/tasks/split_key.rb
```diff
@@ -5,68 +5,51 @@ module Tasks
...
-        pos   = 0
         return [key] if max == 1
 
-        key_parts(key) do |part|
-          parts << part
-          pos += part.length + 1
-          if parts.length + 1 >= max
-            parts << key[pos..] unless pos == key.length
-            break
+        parts = []
```
glebm, line 22: This does a lot more allocations than the previous version.
`split_key` is called a lot, what's the impact on  performance?

tom-brouwer-bex: It's actually faster on my machine (M1 Macbook Air):

On commit [0b4b483c82664f26c5696fb0f6aa1297356e4683](https://github.com/glebm/i18n-tasks/commit/0b4b483c82664f26c5696fb0f6aa1297356e4683) (latest main):
```
> Benchmark.measure { 1_000_000.times do ; I18n::Tasks::SplitKey.split_key('a.#{b.c}.d.<e.f>'); end }
#<Benchmark::Tms:0x0000000105396880 @cstime=0.0, @cutime=0.0, @label="", @real=4.406474999999773, @stime=0.005112000000000005, @total=4.400721000000001, @utime=4.395609>
```

On commit [61bc4fc4c25632798401599f6be6b7a45fef9945](https://github.com/glebm/i18n-tasks/commit/61bc4fc4c25632798401599f6be6b7a45fef9945) (this PR):
```
> Benchmark.measure { 1_000_000.times do ; I18n::Tasks::SplitKey.split_key('a.#{b.c}.d.<e.f>'); end }
#<Benchmark::Tms:0x0000000103042190 @cstime=0.0, @cutime=0.0, @label="", @real=3.340463999998974, @stime=0.005928000000000003, @total=3.336179, @utime=3.330251>
```

glebm: Nice!

Issue #286, opened 2018-04-11
Invalid Value (EasyTranslate::EasyTranslateException)


atd:
When trying to `i18n-tasks translate-missing` with locales with region (like `es-ES` or `en-US`), the following error rises:

```
/usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/easy_translate-0.5.1/lib/easy_translate/request.rb:48:in `perform_raw': Invalid Value (EasyTranslate::EasyTranslateException)
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/easy_translate-0.5.1/lib/easy_translate/translation.rb:34:in `request_translations'
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/easy_translate-0.5.1/lib/easy_translate/threadable.rb:24:in `block in threaded_process'
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/easy_translate-0.5.1/lib/easy_translate/threadable.rb:24:in `map'
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/easy_translate-0.5.1/lib/easy_translate/threadable.rb:24:in `threaded_process'
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/easy_translate-0.5.1/lib/easy_translate/translation.rb:20:in `translate'
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/i18n-tasks-0.9.21/lib/i18n/tasks/google_translation.rb:40:in `fetch_google_translations'
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/i18n-tasks-0.9.21/lib/i18n/tasks/google_translation.rb:30:in `block in google_translate_list'
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/i18n-tasks-0.9.21/lib/i18n/tasks/google_translation.rb:29:in `each'
        from /usr/local/rvm/gems/ruby-2.3.4@ideas4all-brasaas/gems/i18n-tasks-0.9.21/lib/i18n/tasks/google_translation.rb:29:in `map'
...
```

This is because google does not support those locales: https://cloud.google.com/translate/docs/languages

They should be converted to locales without region before calling to `EasyTranslate`

Issue #557, opened 2024-03-29
Google Translate SUPPORTED_LOCALES_WITH_REGION does not handle supported aliases


jqr:
If your app supports any Chinese Traditional locale which isn't `zh-TW`, then `translate-missing` will erroneously fill in Simplified Chinese instead of Traditional Chinese. Most prominently with `zh-Hans` and `zh-HK`.

This is caused by the hardcoded [check and modification](https://github.com/glebm/i18n-tasks/blob/56f378ac102b18c7de2cbcb69cdf2483b0d9f48a/lib/i18n/tasks/translators/google_translator.rb#L46-L51) of locale.

So that locale modification actually hides existing functionality of the Google Translate API to map these into the correct language. Here's a visually distinct word across several Chinese locales to demonstrate API support:

```ruby
EasyTranslate.translate("horse", to: "zh")       # => "马"  # officially supported
EasyTranslate.translate("horse", to: "zh-Hans")  # => "马"  # aliased
EasyTranslate.translate("horse", to: "zh-CN")    # => "马"  # officially supported
EasyTranslate.translate("horse", to: "zh-Hant")  # => "馬"  # aliased
EasyTranslate.translate("horse", to: "zh-HK")    # => "馬"  # aliased
EasyTranslate.translate("horse", to: "zh-TW")    # => "馬"  # officially supported
```

So Google Translate knows which of these `language-REGION` and `language-Script` locales go to Simplified and Traditional, but the current check and modification only allows `zh-TW` to map to Traditional Chinese and everything else goes to Simplified Chinese.

Example: `zh-HK` becomes `zh` because `zh-HK` is not in `SUPPORTED_LOCALES_WITH_REGION`, and then will incorrectly translate text into Simplified Chinese instead of Traditional Chinese.

BTW, I haven't been able to find an official list of all supported aliases, and the [Google Translate API Languages page](https://cloud.google.com/translate/docs/languages) doesn't list any of them.

Issue #556, opened 2024-03-29
Google Translate does case sensitive locale comparison


jqr:
Calling translate-missing with `zh-tw` incorrectly converts to `zh` even though `zh-TW` is known supported. Locales should instead be case-insensitive as per [RFC5646 2.1.1](https://datatracker.ietf.org/doc/html/rfc5646#section-2.1.1) 
> At all times, language tags and their subtags, including private use and extensions, are to be treated as case insensitive: there exist conventions for the capitalization of some of the subtags, but these MUST NOT be taken to carry meaning.
>
> Thus, the tag "mn-Cyrl-MN" is not distinct from "MN-cYRL-mn" or "mN-cYrL-Mn" (or any other combination), and each of these variations conveys the same meaning: Mongolian written in the Cyrillic script as used in Mongolia.

---

```
> i18n-tasks translate-missing --locales zh-tw
```

[This method](https://github.com/glebm/i18n-tasks/blob/56f378ac102b18c7de2cbcb69cdf2483b0d9f48a/lib/i18n/tasks/translators/google_translator.rb#L46-L51) will do a comparison to see if `zh-tw` is in `SUPPORTED_LOCALES_WITH_REGION` which it is not, even though `zh-TW` is. Then it will change from `zh-tw`(Traditional Chinese) to just `zh`(Simplified Chinese). This actually changes the script of translation and is essentially a different language.

If the user had typed `zh-TW` instead, the correct translation would have taken place.

Issue #555, opened 2024-03-29
Google Translate SUPPORTED_LOCALES_WITH_REGION is out of date


jqr:
The [list of languages supported by Google Translate API](https://cloud.google.com/translate/docs/languages) now explicitly supports `mni-Mtei` which is technically a `language-Script`, but functions identically in [this scenario](https://github.com/glebm/i18n-tasks/blob/56f378ac102b18c7de2cbcb69cdf2483b0d9f48a/lib/i18n/tasks/translators/google_translator.rb#L46-L51).

PR #558, opened 2024-03-29
Stop modifying locale before sending to Google Translate

jqr:
In https://github.com/glebm/i18n-tasks/pull/292 the from/to locales were modified to be limited to just the language portion of a locale tag except if it was within one of the listed set of `langague-REGION` tags. This was to fix https://github.com/glebm/i18n-tasks/issues/286.

That change no longer seems to be required by the underlying API:

```ruby
EasyTranslate.translate("horse", to: "es")       # => "caballo"
EasyTranslate.translate("horse", to: "es-es")    # => "caballo"
EasyTranslate.translate("horse", to: "es-ES")    # => "caballo"

EasyTranslate.translate("caballo", to: "en")     # => "horse"
EasyTranslate.translate("caballo", to: "en-us")  # => "horse"
EasyTranslate.translate("caballo", to: "en-US")  # => "horse"
```

And maintaining that list directly in code causes several small issues, #555 #556, and #557 which actually is translating things into the incorrect language altogether.

Given that, it seems worthy of removing the vestigial locale modification code entirely and just relying on new improved behavior of the Google Translate API and EasyTranslate.

### DIFF
--- a/lib/i18n/tasks/translators/google_translator.rb
+++ b/lib/i18n/tasks/translators/google_translator.rb
@@ -22,8 +22,8 @@ module I18n::Tasks::Translators
     def options_for_translate_values(from:, to:, **options)
       options.merge(
         api_key: api_key,
-        from: to_google_translate_compatible_locale(from),
-        to: to_google_translate_compatible_locale(to)
+        from: from,
+        to: to,
       )
     end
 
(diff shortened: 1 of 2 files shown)


glebm:
Thanks!

PR #565, opened 2024-05-07
PrismScanner: Contextual parsing for Rails

davidwessman:
- Adding a scanner based on [Prism](https://github.com/ruby/prism).
- It supports
  - `before_action` in controllers
  - translations in nested method calls
  - `model_name.human`
  - `human_attribute_name`
  
## Note to reviewers

The scanner could replace all our calls to the `parser` gem today and could in the future also replace our matchers.
The harder part is to make it `extendable`, should we allow the end user to implement their own `Prism::Visitor` that would support some new use case? 

I left the email subjects and ViewComponent for a smaller future PR, but I think this is a nice base to start testing it out on some real apps and evaluate if this is something we want to move towards.

----------

## TODO 
- [x] Handle comments
- [x] Add magic comment to skip parsing file
- [x] Handle all arguments to translation method
- [x] Handle Rails-model calls
- [x] Handle more before_actions, or warn that they cannot be processed

## Future features

- [ ] Handle email subjects
- [ ] Handle ViewComponent translations
- [ ] Handle `default` argument


glebm:
We already have a Ruby parser dependency, consider reusing it rather than adding another one.

davidwessman:
> We already have a Ruby parser dependency, consider reusing it rather than adding another one.

Prism is a standard ruby gem and is included by default from Ruby 3.3 and can be installed from Ruby 3.0.
I think we should investigate moving away from parser towards it.

I will do this experiment in Prism and then I can rewrite it to use the old parser.

When parsing controllers I would like to return a key or a fallback, e.g. if there is a relative key in a method name I would like the key with either the method name included or with just the controller namespace.

```ruby
class EventsController < ApplicationController
  def show
    t(".relative")
  end
end
```

Then I would like it to first check for `events.show.relative`, but if that does not exist, fallback to `events.relative`.
Do you have any idea where to start with adding such behaviour?

glebm:
Looks like Prism on JRuby is still a work-in-progress: http://blog.jruby.org/2024/02/jruby-prism-parser

jaredmoody:
+1 for Prism! The speed gains rubocop is getting from using it are 👀

davidwessman:
> Looks like Prism on JRuby is still a work-in-progress: http://blog.jruby.org/2024/02/jruby-prism-parser

I think this means that JRuby are working on _replacing_ their parser with Prism, for their compilation.
Prism can still handle all syntax that JRuby supports and can run in JRuby, so for our use-case it would be okay.

lib/i18n/tasks/scanners/prism_scanners/visitor.rb
```diff
@@ -0,0 +1,168 @@
+# frozen_string_literal: true
+
+require 'prism/visitor'
+require_relative 'nodes'
+
+# Implementation of Prism::Visitor (https://ruby.github.io/prism/rb/Prism/Visitor.html)
+# It processes the parsed AST from Prism and creates a new AST with the nodes defined in prism_scanners/nodes.rb
+# The only argument it receives is comments, which can be used for magic comments.
+# It defines processing of arguments in a way that is neede for the translation calls.
```
davidwessman, line 9:
```suggestion
# It defines processing of arguments in a way that is needed for the translation calls.
```

spec/prism_scanner_spec.rb
```diff
@@ -0,0 +1,420 @@
...
+          Event.model_name.human
+        RUBY
+
+        occurrences = process_string('app/models/event.rb', source)
+
+        expect(occurrences.map(&:first)).to match_array(
+          %w[
+            activerecord.attributes.event.title
+            activerecord.models.event.one
+            activerecord.models.event.other
```
davidwessman, line 202: Maybe we would like some different way to return a `pluralized` translation, e.g. that it would just return `activerecord.models.event` with some `pluralized` setting and then this can be included in the tree.

davidwessman:
@glebm Now I feel like I am ready for a review of this as a _beta_-feature.

I am sorry for a large PR, please let me know if you have any suggestions for making it smaller and more reviewable.

spec/prism_scanner_spec.rb
glebm: Yeah, I think that would be necessary to support pluralized keys properly.
Can be done in a follow-up PR

glebm (approved):

PR #543, opened 2023-12-30
Only uses AST-parser for .html.erb

davidwessman:
- Only uses ErbAst-parser for html files
- Fixes #537

Issue #537, opened 2023-12-03
Inconsistent and unrelated error messages when modifying code


gregorbg:
I experienced a bug when working on a big feature upgrade in our codebase, and I haven't managed to debug it even after 6+ hours.
From what I can tell, it looks like one of the weirdest bugs that I have ever encountered in our code ever, but I will try my best to describe.

## Situation

We are a volunteer-driven association with a big, historically grown codebase. As we are growing, we are aiming to modernise our website code and replace old, wild and ugly Ruby templates (that have grown complicated over time) with React code. So a common use case in our codebase is to remove some `app/views/**/*.html.erb` files and replace them with js(x) files in `app/webpacker`.

## The crash

While working on a feature replacement of a particularly big page, we deleted the old `_competition_form.html.erb`. This leads to a crash of our i18n CI check (called via `I18n::Tasks::BaseTask.new.missing_keys(locales: [:en])`) with the following error message:

```
I18n::Tasks::CommandError:
       Error scanning app/webpacker/lib/requests/routes.js.erb: Cannot resolve relative key ".pending_claim_confirmation_html".
       Set search.relative_roots in config/i18n-tasks.yml (currently ["app/controllers", "app/helpers", "app/mailers", "app/presenters", "app/views"])
```

What's so weird about this:
- The file that is being deleted contains no mention of `.pending_claim_confirmation_html`
- The file in the error message (`app/webpacker/lib/requests/routes.js.erb`) is not related to the file that is being deleted at all.

## Attempts at fixing

The first idea was to simply ignore the `routes.js.erb` mentioned in the error message. Doing so makes the scan proceed without other errors, but reports false positives:

```
Missing 1 i18n keys
en {:key_filter=>nil}
  users
    index
      pending_claim_confirmation_html:  {:occurrences=>[Occurrence(app/views/users/index.html.erb:29, line_pos: 8, pos: 1358, raw_key: .pending_claim_confirmation_html, default_arg: , line:         <th class="email" data-field="email" data-sortable="true">Email</th>)], :type=>:missing_used}
You can also run `i18n-tasks missing -l en' to show them
```

The key that is reported "missing" is the same as in the crash error message above. However, it does not occur in the file `app/views/users/index.html.erb` at all! (The source code snippet from the above output is accurate)

[...]

PR #573, opened 2024-05-13
Uses AST-scanner for all ERB

### LINKED ISSUE #572: Version 1.0.14 thinks any method ending in `_t` is a translation method
I have a method `theme_t` that takes a string and is used in my ERB files. With version 1.0.14, i18n-tasks has started complaining that these string arguments cannot be found in the translation files.

It seems the relevant change is e939d0e1f51baaa5d8f229f54bc18d3cc438d71b, since the changed regex matches `theme_t "foo"` but the original regex did not.

davidwessman:
- Also fixes regex in PatternScanner to handle `theme_t "foo"` being
  marked as a translation.
- Fixes #572

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,88 +1,88 @@
 ## Unreleased
 
-- Uses AST-parser for all ERB-files, not just `.html.erb`
-- [Fixed regex in `PatternScanner`] (https://github.com/glebm/i18n-tasks/issues/572)
+* Uses AST-parser for all ERB-files, not just `.html.erb`
+* [Fixed regex in `PatternScanner`] (https://github.com/glebm/i18n-tasks/issues/572)
 
 ## v1.0.14
 
-- Newlines are now preserved when using Google Translate.
+* Newlines are now preserved when using Google Translate.
   [#567](https://github.com/glebm/i18n-tasks/pull/567)
-- Improved locale name handling for Google Translate.
+* Improved locale name handling for Google Translate.
   [#558](https://github.com/glebm/i18n-tasks/pull/558)
-- Fixes compatibility with some versions of the i18n gem.
+* Fixes compatibility with some versions of the i18n gem.
   [#553](https://github.com/glebm/i18n-tasks/pull/553)
-- Added `i18n-tasks cp` command.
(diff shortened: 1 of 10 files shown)


davidwessman:
@mvz Can you test it out? 🙂

CHANGES.md
```diff
@@ -1,83 +1,88 @@
+## Unreleased
+
+- Uses AST-parser for all ERB-files, not just `.html.erb`
+- [Fixed regex in `PatternScanner`] (https://github.com/glebm/i18n-tasks/issues/572)
+
 ## v1.0.14
 
-* Newlines are now preserved when using Google Translate.
+- Newlines are now preserved when using Google Translate.
```
glebm, line 8: Can you please revert the changes from `-` to `*`?

glebm:
Looks good to me other than the markdown formatting change

davidwessman:
> Looks good to me other than the markdown formatting change

Need to find some better settings for the VS Code repo to avoid automatic formatting.

mvz:
@davidwessman I have tested with these changes and it works now. Thanks!

gregorbg:
I'm a little worried that this may cause regressions for https://github.com/glebm/i18n-tasks/issues/537.

The limitation to only `.html.erb` was deliberately introduced here: https://github.com/glebm/i18n-tasks/pull/543
This current PR looks like it exactly reverts the fix from that other PR. What will happen to `.js.erb` files again when this change is released?

davidwessman:
> I'm a little worried that this may cause regressions for https://github.com/glebm/i18n-tasks/issues/537.
> 
> 
> 
> The limitation to only `.html.erb` was deliberately introduced here: https://github.com/glebm/i18n-tasks/pull/543
> 
> This current PR looks like it exactly reverts the fix from that other PR. What will happen to `.js.erb` files again when this change is released?

There is a new scanner that handles it 🙂 and there are tests for it.

davidwessman:
But please add some tests if you think there needs to be more test coverage for it.

davidwessman:
https://github.com/glebm/i18n-tasks/pull/545

Here is the new parser 🙂

Issue #575, opened 2024-05-18
Configure translation backend in i18n-tasks.yml


jaredmoody:
It would be nice to be able to specify a default translation backend via yaml so I don't have to type `--backend openai` every time.

e.g.

```
translation:
  backend: openai
```

I can submit a PR for this if you like the idea!

glebm:
PRs are welcome! 👍

PR #580, opened 2024-05-24
Allow configuring translation backend in yaml

jaredmoody:
I also tidied up the README to be a little more concise and added examples of backend configuration under each translation backend configuration section.

LMK if you would like any changes.

### DIFF
--- a/lib/i18n/tasks/command/commands/missing.rb
+++ b/lib/i18n/tasks/command/commands/missing.rb
@@ -48,7 +48,9 @@ module I18n::Tasks
             pattern_re = i18n.compile_key_pattern(opt[:pattern])
             missing.select_keys! { |full_key, _node| full_key =~ pattern_re }
           end
-          translated = i18n.translate_forest missing, from: opt[:from], backend: opt[:backend].to_sym
+
+          backend = opt[:backend].presence || i18n.translation_config[:backend]
+          translated = i18n.translate_forest missing, from: opt[:from], backend: backend.to_sym
           i18n.data.merge! translated
           log_stderr t('i18n_tasks.translate_missing.translated', count: translated.leaves.count)
           print_forest translated, opt
--- a/lib/i18n/tasks/configuration.rb
+++ b/lib/i18n/tasks/configuration.rb
@@ -5,7 +5,8 @@ module I18n::Tasks::Configuration # rubocop:disable Metrics/ModuleLength
     base_locale: 'en',
     internal_locale: 'en',
     search: ::I18n::Tasks::UsedKeys::SEARCH_DEFAULTS,
-    data: ::I18n::Tasks::Data::DATA_DEFAULTS
+    data: ::I18n::Tasks::Data::DATA_DEFAULTS,
+    translation_backend: :google
   }.freeze
 
   # i18n-tasks config (defaults + config/i18n-tasks.yml)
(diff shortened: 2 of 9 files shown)


davidwessman:
Fixes #575

lib/i18n/tasks/command/commands/missing.rb
```diff
@@ -48,7 +48,9 @@ def translate_missing(opt = {})
             pattern_re = i18n.compile_key_pattern(opt[:pattern])
             missing.select_keys! { |full_key, _node| full_key =~ pattern_re }
           end
-          translated = i18n.translate_forest missing, from: opt[:from], backend: opt[:backend].to_sym
+
+          backend = i18n.config.dig(:translation, :backend) || opt[:backend]
```
glebm, line 52: This looks like it ignores the command-line flag if the config setting is present?

jaredmoody: Yes, I think this should be the other way around and the CLI flag should override, will revise.

jaredmoody: Ah, I realized why I did this - `opt[:backend]` is always present, because a default is being set to `:google`, so it's not possible to tell here if the CLI flag was provided or it's falling back to the default.

I don't think having a default backend is the best idea; if no backend is specified, getting an error message that requests a Google API Key is strange given the user has not requested to use Google.

I will submit a PR to remove the google default, and then it will be possible to swap the order here. Does that sound good?

jaredmoody:
Just posted #581 which is a prerequisite to this PR.

lib/i18n/tasks/command/commands/missing.rb
glebm: The backend should default to google for backwards compatibility.

Let's set the argument to `nil`. Then here we can do

```ruby
          backend = opt[:backend] || i18n.translation_config[:backend]
```

And set the default backend here:

https://github.com/glebm/i18n-tasks/blob/9852a3f49661babdba58cec841538e637da9f48e/lib/i18n/tasks/configuration.rb#L65

jaredmoody: Sorry, I may not be clear on what you want here, but if by set the argument to `nil` you mean this:

```ruby
arg :translation_backend,
            '-b',
            '--backend BACKEND',
            t('i18n_tasks.cmd.args.desc.translation_backend'),
            parser: OptionParsers::Locale::Parser,
            default: nil # SET TO NIL HERE
```            

This doesn't work because the `OptionParsers::Locale::Parser` defaults to `en` when the argument is nil. This is one reason why in #581 I replaced this with the `Enum::Parser` - which I modified to accept an argument to not default to the first option.

You could re-open that PR and I could modify it to fall back to `:google` for backwards compatibility if you like?

Or, if I'm not understanding, can you clarify what you would like done here?

I'm also fine with you modifying this PR to what you want if that's easier.

glebm: Yeah, makes sense.
Let's do everything in this PR, makes it easier to review and refer to in the release notes.

jaredmoody:
By placing the backend default in the suggested location, there's now a `Metrics/AbcSize` violation on that method. LMK how you would like to resolve.

glebm:
That's fine, please disable `Metrics/AbcSize` for that method

glebm:
Looks good to me! Disable the rubocop warning and I'll merge :+1:

lib/i18n/tasks/configuration.rb
```diff
@@ -62,6 +63,7 @@ def data_config
   def translation_config
```
davidwessman, line 63:
```suggestion
  def translation_config # rubocop:disable Metrics/AbcSize
```

PR #581, opened 2024-05-28
Remove default translation backend

jaredmoody:
Modify the Enum option parser to have an option, `default` which allows indicating if the option parser should fall back to the first valid value.

For translation backends, we do not want to have a default because the user must configure the translation backend.

Modify the translation_backend argument to provide an error message when no backend is specified

### DIFF
--- a/lib/i18n/tasks/translation.rb
+++ b/lib/i18n/tasks/translation.rb
@@ -11,7 +11,7 @@ module I18n::Tasks
     # @param [String] from locale
     # @param [:deepl, :openai, :google, :yandex] backend
     # @return [I18n::Tasks::Tree::Siblings] translated forest
-    def translate_forest(forest, from:, backend: :google)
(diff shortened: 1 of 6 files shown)


lib/i18n/tasks/translation.rb
```diff
@@ -22,7 +22,7 @@ def translate_forest(forest, from:, backend: :google)
       when :yandex
         Translators::YandexTranslator.new(self).translate_forest(forest, from)
       else
-        fail CommandError, "invalid backend: #{backend}"
+        fail CommandError, backend.present? ? "invalid backend: #{backend}" : 'no backend specified'
```
jaredmoody, line 25: AFAICT this isn't translated because it should never happen - the option parser should fail before this line is reached

glebm:
As mentioned in the other PR, let's keep `google` as default for backwards compatibility.

PR #583, opened 2024-06-07
Provide a 'Changelog' link on rubygems.org/gems/i18n-tasks

mark-young-atg:
By providing a 'changelog_uri' in the metadata of the gemspec a 'Changelog' link will be shown on https://rubygems.org/gems/i18n-tasks which makes it quick and easy for someone to check on the changes introduced with a new version.

Details of this functionality can be found on https://guides.rubygems.org/specification-reference/

### DIFF
--- a/i18n-tasks.gemspec
+++ b/i18n-tasks.gemspec
@@ -27,6 +27,7 @@ Gem::Specification.new do |s|
   TEXT
   s.homepage = 'https://github.com/glebm/i18n-tasks'
   s.metadata = {
+    'changelog_uri' => 'https://github.com/glebm/i18n-tasks/blob/main/CHANGES.md',
     'issue_tracker' => 'https://github.com/glebm/i18n-tasks',
     'rubygems_mfa_required' => 'true'
   }
--- a/i18n-tasks.gemspec
+++ b/i18n-tasks.gemspec
@@ -29,7 +29,8 @@ Gem::Specification.new do |s|
   s.metadata = {
     'changelog_uri' => 'https://github.com/glebm/i18n-tasks/blob/main/CHANGES.md',
     'issue_tracker' => 'https://github.com/glebm/i18n-tasks',
-    'rubygems_mfa_required' => 'true'
+    'rubygems_mfa_required' => 'true',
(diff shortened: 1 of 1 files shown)


davidwessman (approved):
Ah, that is a great idea 🙂

i18n-tasks.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |s|
   TEXT
   s.homepage = 'https://github.com/glebm/i18n-tasks'
   s.metadata = {
+    'changelog_uri' => 'https://github.com/glebm/i18n-tasks/blob/main/CHANGES.md',
     'issue_tracker' => 'https://github.com/glebm/i18n-tasks',
```
davidwessman, line 31: Could you update it based on the documentation as well?
```suggestion
    'bug_tracker_uri' => 'https://github.com/glebm/i18n-tasks',
```

davidwessman: And maybe we can add `source_code_uri` to point to the repo as well?

mark-young-atg:
@davidwessman thank you. I've just done a bit of digging and I think `issue_tracker` was valid with [gemspec 2.4](https://docs.ruby-lang.org/en/2.4.0/Gem/Specification.html) and with [2.5](https://docs.ruby-lang.org/en/2.5.0/Gem/Specification.html) it changed to `bug_tracker_uri`

Interestingly the [rubygems page](https://rubygems.org/gems/i18n-tasks) already has a `Bug Tracker` link which goes to "https://github.com/glebm/i18n-tasks/issues" despite that not being explicitly specified in the metadata. I'm inclined not to touch it. "If it ain't broke...". I've failed to turn up any documentation that talks about `issue_tracker` being deprecated or that it automatically adds "/issues" to the provided URL if needed.

I've added a commit with the source_code_uri

i18n-tasks.gemspec
mark-young-atg: 👍 Done in [9732435](https://github.com/glebm/i18n-tasks/pull/583/commits/97324354474c1638214a4dc97bc8ecb881f267c7)

davidwessman (approved):

Issue #574, opened 2024-05-14
How to force retranslate from base


lorismaz:
Hi,

Is there a way to force translate in all other locales from base if we change the content of a base translation ?

For example, with base being en,  

`cart: Cart`

Change to 

`cart: Basket`

I want to force retranslate all destinations locales other than base. 

Thanks!

PR #585, opened 2024-06-16
Adds command to translate a pattern of keys

davidwessman:
- I cannot figure out how to select the data properly, can you help me?
- Related to #574

### DIFF
--- a/lib/i18n/tasks/command/commands/translate.rb
+++ b/lib/i18n/tasks/command/commands/translate.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module I18n::Tasks
+  module Command
+    module Commands
+      module Translate
+        include Command::Collection
+        include I18n::Tasks::KeyPatternMatching
+
+        cmd :translate,
+            pos: '[pattern]',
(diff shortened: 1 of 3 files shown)


lib/i18n/tasks/command/commands/translate.rb
```diff
@@ -0,0 +1,33 @@
...
+        include Command::Collection
+        include I18n::Tasks::KeyPatternMatching
+
+        cmd :translate,
+            pos: '[pattern]',
+            desc: t('i18n_tasks.cmd.desc.translate'),
+            args: [:locale, :locale_to_translate_from, arg(:data_format).from(1), :translation_backend]
+
+        def translate(opts = {})
+          forest = i18n.tree(opts[:from])
```
davidwessman, line 16: @glebm Do you have a suggestion for creating a tree with all keys from `opts[:from]` locale to send into the `translate_forest` method? I have tried a few ways but cannot figure out the correct format.

glebm: The forest root nodes should be the locales to translate to.

For example, `{"en": {"key": "Value"}}`

PR #592, opened 2024-07-24
update i18n tasks interpolations for deepl

barak-neeman:
i18n-tasks normally send DeepL something like:
Taak "<i18n>%{todo}</i18n>" toegewezen aan <i18n>%{user}</i18n>
and gets back
<i18n>%{todo}</i18n>To-dos " " zugewiesen an <i18n>%{user}</i18n>
That mess is the reason why this PR is needed

In this PR we send a custom <var> tag like:
Taak "<var stache="%{todo}" sub="RYX">RYX</var>" toegewezen aan <var stache="%{user}" sub="QFN">QFN</var>
and get back
To-dos "<var stache="%{todo}" sub="RYX">RYX</var>" zugewiesen an <var stache="%{user}" sub="QFN">QFN</var>

XML attributes aren't modified by DeepL, so we can rely on stache (short for moustache) to come back with the original interpolation key, and sub to come back with the text we substituted for the interpolation key.

Now we can allow DeepL to "read" the <var> tags (whereas <i18n> were ignored), and that makes for better translations, and solves the problem of interpolations getting pushed to the front.

What we get back isn't all sunshine and rainbows, though, DeepL will still occasionally mangle it. More details in the code.

### DIFF
--- a/lib/i18n/tasks/interpolations.rb
+++ b/lib/i18n/tasks/interpolations.rb
@@ -3,25 +3,30 @@
 module I18n::Tasks
   module Interpolations
     class << self
-      attr_accessor :variable_regex
+      attr_accessor :variable_regex, :tag_pairs, :tag_with_localized_value_regex
     end
-    @variable_regex = /(?<!%)%{[^}]+}/.freeze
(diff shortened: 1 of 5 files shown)


barak-neeman:
Wrong repo, my mistake, sorry.
This isn't ready yet

Issue #582, opened 2024-05-30
Feature: Translation of few/many keys based on the locale


marselmustafin:
I have such keys in my base `en` locale:
```
    friends:
      one: one friend
      other: "${count} friends"
      zero: no friends
```

When I translate it to other languages with `translate-missing` I have the same keys everywhere. E.g. in Russian:
```
    friends:
      one: один друг
      other: "%{count} друзей"
      zero: нет друзей
```

But I would like to have more values included in locales that require more pluralization keys, expected result for Russian:

```
    friends:
      one: один друг
      few: "%{count} друга"
      many: "%{count} друзей"
      other: "%{count} друзей"
      zero: нет друзей
```

Is it possible to achieve this result currently? Or it requires additional feature implementation?

marselmustafin:
@davidwessman no, I never faced with such ready-made API. We have been using Lokalise for a long time and they provided us with these keys, but afaik they use real people for that.

In `i18n-tasks` project there's a `plural` mode exists for `missing` task which shows all missed plural keys for non-en locales like `ar`, `ru`, etc. But `translate-missing` provides only 1-to-1 keys translation. Even adding of `few`/`many` keys to en didn't help because the translation engine `google`/`openai` is not aware of context of the number and it gives wrong translation. 

I think it would be awesome to upgrade `translate-missing` task to add translations for plurals as well. As an idea we could pass some number related to the key like `few`: 2, `many`: 5 to the translation request and then replace it with `%{count}` again.

davidwessman:
Yes, that would be interesting to try out. Send translate for multiple plural-keys and see if they are different

davidwessman:
Seems like it would be pretty complex rules to set this up:
https://guides.rubyonrails.org/i18n.html#locale-specific-rules
https://cldr.unicode.org/index/cldr-spec/plural-rules
https://www.unicode.org/cldr/charts/45/supplemental/language_plural_rules.html

markedmondson:
Ok I think I have this "working" at least as a proof of concept. 

As per the comments above, it's a little tricky and assumptions have to be made - as a result it probably only works well from `en` to other languages because I took the approach of using the `other` key to fill in the missing keys so it would work well with translations. Ultimately, if you're just looking to get the missing keys, that's a lot easier.

I pushed our large application using it through the OpenAI translator and it did a pretty good job at ensuring the correct keys were translated.

I'll push up a branch, there were a bunch of other changes that I made along the way so I'll try and split it up nicely.

PR #596, opened 2024-08-30
Identify and translate missing plural keys

markedmondson:
From https://github.com/glebm/i18n-tasks/issues/582

* For locales identified in a call to `translate` commands, determines the existing and required pluralization keys and adds the nodes to the forest before translating. It will additionally remove unrequired plural translation keys. 
* Adds a label to the front of missing plural keys when running the `missing` command as the table headers were a little unclear.
* Add a --skip-interpolation option to the `missing` command (but specifically intended for translation) that won't send **only** interpolated strings to translate, the reasoning behind this was a) I found OpenAI would end up getting a mismatched number of results returned if there were a lot of them and b) it's kind of wasteful as a correctly configured Rails implementation would fallback and work correctly.

I think there's still some cleanup to do here, it feels a little clobbered together as I was learning as I went so totally happy to solicit feedback (or hear and attempt to address any problems).

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -23,6 +23,7 @@ en:
           nostdin: Do not read from stdin
           out_format: 'Output format: %{valid_text}'
           pattern_router: 'Use pattern router: keys moved per config data.write'
+          skip_interpolation: Skip translating strings that are straight interpolations (eg. "%{comment}")
           strict: >-
             Avoid inferring dynamic key usages such as t("cats.#{cat}.name"). Takes precedence over
             the config setting if set.
@@ -112,6 +113,7 @@ en:
     missing:
       details_title: Value in other locales or source
       none: No translations are missing.
+      plural: Missing plural keys
     openai_translate:
       errors:
         no_api_key: >-
--- a/lib/i18n/tasks/missing_keys.rb
+++ b/lib/i18n/tasks/missing_keys.rb
@@ -19,12 +19,12 @@ module I18n::Tasks
 
     # @param types [:used, :diff, :plural] all if `nil`.
     # @return [Siblings]
-    def missing_keys(locales: nil, types: nil, base_locale: nil)
+    def missing_keys(locales: nil, types: nil, base_locale: nil, skip_interpolation: false)
       locales ||= self.locales
       types   ||= missing_keys_types
       base = base_locale || self.base_locale
       types.inject(empty_forest) do |f, type|
-        f.merge! send(:"missing_#{type}_forest", locales, base)
+        f.merge! send(:"missing_#{type}_forest", locales, base, skip_interpolation)
       end
     end
 
@@ -35,36 +35,48 @@ module I18n::Tasks
       end
     end
 
-    def missing_diff_forest(locales, base = base_locale)
+    def missing_diff_forest(locales, base = base_locale, skip_interpolation = false)
       tree = empty_forest
       # present in base but not locale
       (locales - [base]).each do |locale|
-        tree.merge! missing_diff_tree(locale, base)
+        tree.merge! missing_diff_tree(locale, base, skip_interpolation)
       end
       if locales.include?(base)
         # present in locale but not base
         (self.locales - [base]).each do |locale|
-          tree.merge! missing_diff_tree(base, locale)
+          tree.merge! missing_diff_tree(base, locale, skip_interpolation)
         end
       end
       tree
     end
 
-    def missing_used_forest(locales, _base = base_locale)
+    def missing_used_forest(locales, _base = base_locale, _skip_interpolation = false)
       locales.inject(empty_forest) do |forest, locale|
         forest.merge! missing_used_tree(locale)
       end
     end
 
-    def missing_plural_forest(locales, _base = base_locale)
+    def missing_plural_forest(locales, base = base_locale, _skip_interpolation = false)
       locales.each_with_object(empty_forest) do |locale, forest|
         required_keys = required_plural_keys_for_locale(locale)
         next if required_keys.empty?
 
         tree = empty_forest
-        plural_nodes data[locale] do |node|
+        plural_nodes data[base] do |node|
           children = node.children
-          present_keys = Set.new(children.map { |c| c.key.to_sym })
+
+          # Get the existing translated node if it exists
+          translated_node = data[locale].get("#{locale}." + node.full_key(root: false))
+          present_keys = if translated_node
+                           # If it's a single existing (non-plural) translation, skip it
+                           next unless translated_node.value.nil?
+
+                           # Otherwise get the existing plural keys
+                           Set.new(translated_node.children.map { |c| c.key.to_sym })
+                         else
+                           Set.new
+                         end
+          # Compare the keys to those existing in base
           next if ignore_key?(node.full_key(root: false), :missing)
           next if present_keys.superset?(required_keys)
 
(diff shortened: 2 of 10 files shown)


config/locales/en.yml
```diff
@@ -23,6 +23,7 @@ en:
           nostdin: Do not read from stdin
           out_format: 'Output format: %{valid_text}'
           pattern_router: 'Use pattern router: keys moved per config data.write'
+          skip_interpolation: Skip translating strings that are straight interpolations (eg. "%{comment}")
```
glebm, line 26: Rather than adding an option, perhaps we should always simply copy such keys verbatim? There is never a reason to translate them.

lib/i18n/tasks/missing_keys.rb
```diff
@@ -35,36 +35,48 @@ def eq_base_keys(opts = {})
...
         next if required_keys.empty?
 
         tree = empty_forest
-        plural_nodes data[locale] do |node|
+        plural_nodes data[base] do |node|
           children = node.children
-          present_keys = Set.new(children.map { |c| c.key.to_sym })
+
+          # Get the existing translated node if it exists
+          translated_node = data[locale].get("#{locale}." + node.full_key(root: false))
```
glebm, line 69: I think this is the same as:

```suggestion
          translated_node = data[locale].get(node.full_key(root: true))
```

lib/i18n/tasks/translators/base_translator.rb
```diff
@@ -102,6 +133,10 @@ def parse_value(untranslated, each_translated, opts)
         else
           untranslated
         end
+      rescue StopIteration => e
+        require 'pry'
```
glebm, line 137: Left-over debugging code?

lib/i18n/tasks/missing_keys.rb
markedmondson: Let me double check but I think the goal here is to to copy the value from the locale we're iterating over from the passed locales in L60 (if it exists). If we used the node value, it'd give us the value from the base locale each time.

config/locales/en.yml
markedmondson: That would be preferential in our use case but I didn't want to assume...

spec/translator_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      missing = task.missing_plural_forest(['ru'], 'en')
+      result = task.translate_forest(missing, from: 'en', backend: :test)
+
+      expect(result.to_hash).to eq(ru_hash)
+    end
+  end
+
+  describe 'multi-line' do
+
+  end
```
markedmondson, line 115: TODO: Remove

lib/i18n/tasks/translators/base_translator.rb
```diff
@@ -51,7 +82,7 @@ def fetch_translations(list, opts)
       # @param [Array<[String, Object]>] list of key-value pairs
       # @return [Array<String>] values for translation extracted from list
       def to_values(list, opts)
-        list.map { |l| dump_value(l[1], opts) }.flatten.compact
+        list.map { |l| dump_value(l[1], opts) }.compact
```
markedmondson, line 85: This could probably go out as a "fix" on it's own - It was compacting array YAML values into the base array which caused the corresponding array size not to match (and also attempted to translate each line as an individual value)

lib/i18n/tasks/translators/base_translator.rb
markedmondson: Totally :)

Interestingly I was hitting it quite a bit using OpenAI when we had a batch of translations that was largely straight interpolations because it would return a mismatch of values, eg, we'd send 50 translation strings and get 49 back. Even specifically prompting to give the matching amount back, I guess ChatGPT is just bad at counting.

lib/i18n/tasks/missing_keys.rb
```diff
@@ -93,9 +109,14 @@ def load_rails_i18n_pluralization!(locale)
     end
 
     # keys present in compared_to, but not in locale
-    def missing_diff_tree(locale, compared_to = base_locale)
-      data[compared_to].select_keys do |key, _node|
-        locale_key_missing? locale, depluralize_key(key, compared_to)
+    def missing_diff_tree(locale, compared_to = base_locale, skip_interpolation = false)
+      data[compared_to].select_keys do |key, node|
+        # Skip if the string is just an interpolation, eg: "%{body}"
+        if skip_interpolation && node.value.to_s.match?(/^%{.*}$/)
```
markedmondson, line 115: TODO: Use `BaseTranslator::INTERPOLATION_KEY_RE`

lib/i18n/tasks/translators/base_translator.rb
markedmondson: Seems like this might actually not be entirely correct.

Our usage is that array values are joined through Slim into paragraphs but obviously there can be cases where an array isn't intended to be used that way. 
We would expect the array to be compacted and translated in its whole context but there's an example in the test files of "encouragement" which doesn't seem to fit this model.

Thoughts on how to proceed?

markedmondson: Got it :)

config/locales/en.yml
markedmondson: Actually we don't even copy them, we have Rails configured with fallback which then just works... Not sure that's a safe assumption either though.

glebm: There is never any reason to translate interpolation-only values, so let's skip them unconditionally.

markedmondson: Sounds good to me! :)

markedmondson:
I spent a little more time on this on a large translation (~610 keys in 30 locales) and added a tweak.

Namely, ChatGPT is pretty bad at returning the correct number of responses back, I've added the array size to the prompt but even then it struggles if there are repeated translations - so there's a fail safe that'll retry and then fall back to doing line by line. I also found dropping the default array size from 50 to 25 improved things somewhat.

markedmondson:
One thing I noted which could perhaps be improved here is that it only adds missing plural keys, it doesn't remove. 

So what seems to happen is we add a new pluralization in en.yml with `one`, `other` and `zero` keys (`zero` being optional). i18n-tasks will translate those 2 (or 3) keys even when the corresponding locale might not need them. This then enforces the pluralization keys that are required (eg `one`, `few`, `many`, `other` in ru.yml) so we'll end up with `one`, `few`, `many`, `other`, and `zero` in ru.yml - the `zero` isn't used, however.

lib/i18n/tasks/base_task.rb
```diff
@@ -39,6 +39,9 @@ class BaseTask
       include Data
       include Stats
 
+      INTERPOLATION_KEY_RE = /%\{[^}]+}/.freeze
+      STRICT_INTERPOLATION_KEY_RE = /^%\{[^}]+\}$/.freeze
```
glebm, line 43: `INTERPOLATION_ONLY_KEY_RE`?

config/locales/en.yml
glebm: Sure, makes sense to have that as an option.

lib/i18n/tasks/missing_keys.rb
glebm: Ah, that makes sense

markedmondson:
> Edit: Looks like there's a problem with the removal of pluralization keys that aren't required, eg we translate one, other and zero from en.yml, they get copied to ru.yml even though it only needs one, few, many, other because it gets caught and merged in the missing_diff_forest (even though missing_plural_forest removes it).

I fixed this in [7e47ded](https://github.com/glebm/i18n-tasks/pull/596/commits/7e47ded817d8909c7ea5464a29047eab86381d4a)

lib/i18n/tasks/base_task.rb
markedmondson: [d15e17c](https://github.com/glebm/i18n-tasks/pull/596/commits/d15e17caa56dabfd7db6dcbe593e500de5b8c262)

lib/i18n/tasks/missing_keys.rb
markedmondson: [f655f32](https://github.com/glebm/i18n-tasks/pull/596/commits/f655f32095f8c9c34c4327e769dfae67234527f4)

PR #598, opened 2024-09-27
feat: add watsonx translation backend

michaelroudnitski:
Hello, really cool gem :)

We are using watsonx to translate our application with llama 3.2. I was inspired by the [openai translator PR](https://github.com/glebm/i18n-tasks/pull/519) and thought i would go ahead and add an alternative backend option with watsonx


glebm:
Thanks! Please fix the rubocop warnings (you can disable the new warnings for `translation_config`) and I'll merge

PR #601, opened 2024-10-20
Adds failing spec

davidwessman:


glebm:
This may be tricky to fix with a regex.

Currently it is defined here:

https://github.com/glebm/i18n-tasks/blob/d0453929603379bea15577065fecfc79a045290e/lib/i18n/tasks/scanners/ruby_key_literals.rb#L5

A possible workaround is to dissallow letters and `]` after the `"` using a negative lookahead, e.g.:

```ruby
LITERAL_RE = /^:?(["']).+?\1(?![\p{L}\]])|:\w+/.freeze
```
https://rubular.com/r/g77VbORrgrqZdx

davidwessman:
Looked at this a bit more, and do not know how we can handle it with regex.
So this will have to be a limitation.

Issue #359, opened 2020-10-13
Find missing keys that contain spaces


bernb:
Task `missing` does not find missing keys that contain spaces, example: `<%= t('some key') %>`. Those are however valid keys (at least for `I18n::Backend::ActiveRecord`) and they get found by the translation backend if translations are added.
Context (layout file):
```
    <div class="row">
      <div class="col-3">
        <dt><%= t('some key') %></dt>
      </div>
      <div class="col">
        <dd><%= @some_output %></dd>
      </div>
    </div>
```

davidwessman:
@bernb What does your translation source look like?

PR #604, opened 2024-10-23
Allow spaces in locale key

Hassan01-dev:
This update allows spaces in translation keys.

Previously, keys containing spaces (e.g., "Key with Spaces") were not considered valid and were removed from the results after scanning the file. With this change, translation keys with spaces are now recognized as valid.

Closes https://github.com/glebm/i18n-tasks/issues/359

### DIFF
--- a/lib/i18n/tasks/scanners/ruby_key_literals.rb
+++ b/lib/i18n/tasks/scanners/ruby_key_literals.rb
@@ -20,7 +20,7 @@ module I18n::Tasks::Scanners
       literal
     end
 
-    VALID_KEY_CHARS = %r{(?:[[:word:]]|[-.?!:;À-ž/])}.freeze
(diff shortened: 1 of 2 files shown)


lib/i18n/tasks/scanners/ruby_key_literals.rb
```diff
@@ -20,7 +20,7 @@ def strip_literal(literal)
       literal
     end
 
-    VALID_KEY_CHARS = %r{(?:[[:word:]]|[-.?!:;À-ž/])}.freeze
+    VALID_KEY_CHARS = %r{(?:[[:word:]]|[-.?!:;À-ž/]|(?<=\w)\s(?=\w))}.freeze
```
glebm, line 23: Should this perhaps match all Unicode letters in the context rather than just `a-z`?
https://rubular.com/r/dOnUgvJhli6Il2

Hassan01-dev: Yes, updated the regex it to match all Unicode letters and not just a-z.

glebm:
Thanks!

PR #605, opened 2024-11-06
[WIP] only match files with configured locales

gr8bit:
fixes too broad locale regexp which previously matched any filename when `%{locale}.yml` was being used in read-section.

Issue yet to be created

### DIFF
--- a/lib/i18n/tasks/data/router/isolating_router.rb
+++ b/lib/i18n/tasks/data/router/isolating_router.rb
@@ -9,11 +9,12 @@ module I18n::Tasks
     class IsolatingRouter
       include ::I18n::Tasks::KeyPatternMatching
 
-      attr_reader :config_read_patterns, :base_locale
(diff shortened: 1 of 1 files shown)


glebm:
>  It requires the user to set the available locales in the i18n-tasks.yml configuration file, at least for the isolating router.

That seems like a very significant requirement change. Can we avoid requiring that somehow?

gr8bit:
> > It requires the user to set the available locales in the i18n-tasks.yml configuration file, at least for the isolating router.
> 
> That seems like a very significant requirement change. Can we avoid requiring that somehow?

In order to keep the locale regex as non-greedy as possible, we could either
1. specify the actual locales to match or
2. make assumptions about the characters used in the locales and compile those into a regex

Maybe for 1. we could take the locales from somewhere else automatically? I don't know the depths of i18n-tasks so any input/help/suggestion would be appreciated. :)

Issue #590, opened 2024-07-15
OpenAI: unexpected token at '```json 


yshmarov:
I get an error when trying to translate with OpenAI

```
Users/yshmarov/.rbenv/versions/3.3.3/lib/ruby/gems/3.3.0/gems/json-2.7.2/lib/json/common.rb:220:in `parse': unexpected token at '```json (JSON::ParserError)
[
  "Wir werden Sie kontaktieren, um Ihnen mitzuteilen, dass Ihre E-Mail-Adresse auf X__0 geändert wurde.",
  "Erwartetes Ergebnis",
  "Eo",
  "ko",
  "Sie können den API-Schlüssel finden",
  "hier",
  "Bezahlt",
  "Haben Sie bereits ein SR-Konto?",
  "SR arbeitet mit Partnern zusammen, um Ihre Bank sicher zu verbinden. Dies sind regulierte Organisationen, die sicherstellen, dass Ihre Daten geschützt sind.\n\nDies entspricht d
```

```
OPENAI_MODEL=gpt-4o OPENAI_API_KEY=sk-***foobar i18n-tasks translate-missing --backend=openai
```

Does anybody else have this issue?

yshmarov:
the above issue seems to be related to `gpt-4o`.

However when I run `gpt-3.5-turbo`, I get another error:

```
 i18n-tasks translate-missing --backend=openai
warning: parser/current is loading parser/ruby33, which recognizes 3.3.4-compliant syntax, but you are running 3.3.0.
Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
#StandWithUkraine
/Users/yaroslavshmarov/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/i18n-tasks-1.0.14/lib/i18n/tasks/translators/base_translator.rb:98:in `next': iteration reached an end (StopIteration)
	from /Users/yaroslavshmarov/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/i18n-tasks-1.0.14/lib/i18n/tasks/translators/base_translator.rb:98:in `parse_value'
	from /Users/yaroslavshmarov/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/i18n-tasks-1.0.14/lib/i18n/tasks/translators/base_translator.rb:90:in `block in parse_value'
[...]

PR #611, opened 2025-01-15
fix(openai_translator): Enforce JSON Response format

ehimen-io:
Hello! Really appreciate this gem for translation tasks!

I get an issue where the GPT models sometimes responds with improperly formatted JSON (as seen in #590), so this PR contains an additional system prompt to prevent this. I also turned on [JSON mode](https://platform.openai.com/docs/guides/structured-outputs#json-mode) to properly structure the output.

Let me know if there are any changes needed to be made. If not, it would be great if we could create a new release of this gem :)

PS: Ignore the branch name, this fix was originally for Watsonx but I changed gears.


glebm:
Thanks!

PR #614, opened 2025-01-25
OpenAI: log errors

navidemad:
```
bundle exec i18n-tasks translate-missing --backend=openai 
bundler: failed to load command: i18n-tasks (/Users/dev/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/bin/i18n-tasks)                                                                                                                           >  ETA: ??:??:?? 0/161 (0%)
/Users/dev/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/faraday-2.12.2/lib/faraday/response/raise_error.rb:30:in 'Faraday::Response::RaiseError#on_complete': the server responded with status 400 (Faraday::BadRequestError)
```

In order to have more detailed informations you can pass now OPENAI_LOG_ERRORS as environment variable to display more informations.

### DIFF
--- a/lib/i18n/tasks/translators/openai_translator.rb
+++ b/lib/i18n/tasks/translators/openai_translator.rb
@@ -52,7 +52,14 @@ module I18n::Tasks::Translators
     private
 
     def translator
-      @translator ||= OpenAI::Client.new(access_token: api_key)
+      @translator ||=
+        if @i18n_tasks.translation_config[:openai_log_errors]
+          OpenAI::Client.new(access_token: api_key, log_errors: @i18n_tasks.translation_config[:openai_log_errors]) do |f|
+            f.response :logger, Logger.new($stdout), bodies: true
+          end
+        else
+          OpenAI::Client.new(access_token: api_key)
+        end
     end
 
     def api_key
--- a/lib/i18n/tasks/translators/openai_translator.rb
+++ b/lib/i18n/tasks/translators/openai_translator.rb
@@ -52,14 +52,12 @@ module I18n::Tasks::Translators
     private
 
     def translator
-      @translator ||=
-        if @i18n_tasks.translation_config[:openai_log_errors]
-          OpenAI::Client.new(access_token: api_key, log_errors: @i18n_tasks.translation_config[:openai_log_errors]) do |f|
-            f.response :logger, Logger.new($stdout), bodies: true
-          end
-        else
-          OpenAI::Client.new(access_token: api_key)
-        end
+      @translator ||= OpenAI::Client.new(
+        access_token: api_key,
+        # Highly recommended in development, so you can see what errors OpenAI is returning.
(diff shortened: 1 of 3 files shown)


dankimio:
Removing `f.response :raise_error` in [ruby-openai](https://github.com/alexrudall/ruby-openai) also helps — [openai/http.rb#L79](https://github.com/alexrudall/ruby-openai/blob/v7.3.1/lib/openai/http.rb#L79.)


```ruby
connection = Faraday.new do |f|
  f.options[:timeout] = @request_timeout
  f.request(:multipart) if multipart
  f.use MiddlewareErrors if @log_errors
  # f.response :raise_error
  f.response :json
end
```

Not sure why they chose to hide log output from errors.

lib/i18n/tasks/translators/openai_translator.rb
```diff
@@ -52,7 +52,14 @@ def no_results_error_message
     private
 
     def translator
-      @translator ||= OpenAI::Client.new(access_token: api_key)
+      @translator ||=
+        if @i18n_tasks.translation_config[:openai_log_errors]
+          OpenAI::Client.new(access_token: api_key, log_errors: @i18n_tasks.translation_config[:openai_log_errors]) do |f|
+            f.response :logger, Logger.new($stdout), bodies: true
```
glebm, line 58: 1. Errors should be logged to `STDERR`
2. I think we can always log the errors, no need for the environment variable

navidemad:
Differents repository are recommanding to set this option only to true for development, as it can leak leak private data to your logs.
So i made it related to RAILS_ENV / RAKE_ENV.

lib/i18n/tasks/translators/openai_translator.rb
```diff
@@ -52,7 +52,12 @@ def no_results_error_message
     private
 
     def translator
-      @translator ||= OpenAI::Client.new(access_token: api_key)
+      @translator ||= OpenAI::Client.new(
+        access_token: api_key,
+        # Highly recommended in development, so you can see what errors OpenAI is returning.
+        # Not recommended in production because it could leak private data to your logs.
+        log_errors: ENV.fetch('RAILS_ENV', ENV.fetch('RACK_ENV', '')) == 'development'
```
glebm, line 59: `RAILS_ENV` and `RACK_ENV` are usually not defined when running i18n-tasks.
While it supports Rails, it does it entirely via static analysis, and does not load rails itself.

It is possible of course to load i18n-tasks as a library from within rails, in which case it'd be good to respect this.

However, most of the time, this string will be empty. So let's also log errors when it's empty.

navidemad: Ok i see. Anyways, in your README.md most of the time the gem will be installed only in the group :development anyways.

glebm (approved):

PR #615, opened 2025-01-25
Append JSON instructions to OpenAI system prompt

navidemad:
In order to fix response_format json_object introduced in the following commit: https://github.com/glebm/i18n-tasks/commit/a0e2fef0b850990490e96facdfff9cb5b394f78f

```
INFO -- : response: {
  "error": {
    "message": "'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'.",
    "type": "invalid_request_error",
    "param": "messages",
    "code": null
  }
}
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* Append JSON instructions to OpenAI system prompt to be able to use response_format json_object
 * Uses AST-parser for all ERB-files, not just `.html.erb`
 * [Fixed regex in `PatternScanner`] (https://github.com/glebm/i18n-tasks/issues/572)
 * Adds contextual parser to support more Rails-translations
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ## Unreleased
 
-* Append JSON instructions to OpenAI system prompt to be able to use response_format json_object
+* Append JSON instructions to OpenAI system prompt to be able to use response_format json_object [#615](https://github.com/glebm/i18n-tasks/pull/615)
 * Uses AST-parser for all ERB-files, not just `.html.erb`
 * [Fixed regex in `PatternScanner`] (https://github.com/glebm/i18n-tasks/issues/572)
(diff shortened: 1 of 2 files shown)


davidwessman:
@navidemad can you fix the linting error? (Add an ignore for the too long line).

@glebm do you have any input on how we should handle these instructions? I have not used OpenAI for any translation.

navidemad:
@davidwessman I fixed the linting error as requested

glebm:
I don't do ruby at all anymore but this PR makes sense to me in principle.

dankimio:
@jaredmoody yes, this looks good. not sure if there's a need for a separate `JSON_FORMAT_INSTRUCTIONS_SYSTEM_PROMPT ` constant. wouldn't it be easier to include that phrase in the `DEFAULT_SYSTEM_PROMPT`? but nice approach with `concat` and already better than what we have currently.

navidemad:
@dankimio  I made it that way because if someone sets the :openai_system_prompt option—for example, to create a custom prompt—but forgets to include this JSON part, it will also fail. This way, it will be appended automatically, preventing this issue.

davidwessman (approved):

jaredmoody:
We do something very similar to this in our prompt, so LGTM.

As a future enhancement though, it would be nice to request JSON from the API by passing `response_format: { type: "json_object" }` so that the result is always JSON and it's not necessary to include that in the prompt.

https://github.com/alexrudall/ruby-openai?tab=readme-ov-file#json-mode

navidemad:
> We do something very similar to this in our prompt, so LGTM.
> 
> As a future enhancement though, it would be nice to request JSON from the API by passing `response_format: { type: "json_object" }` so that the result is always JSON and it's not necessary to include that in the prompt.
> 
> https://github.com/alexrudall/ruby-openai?tab=readme-ov-file#json-mode

https://github.com/glebm/i18n-tasks/blob/main/lib/i18n/tasks/translators/openai_translator.rb#L101
It is already.

PR #618, opened 2025-02-09
PrismScanner: Support more node types

davidwessman:
- There are lots of nodes used in e.g. Rails controllers that needs handling
- Fixes scanning for translations in before_actions with blocks and lambdas


glebm:
Nice!

PR #623, opened 2025-03-02
PrismScanner: Rewrite to avoid implementing visit

davidwessman:
- This rewrites the Visitor to avoid implementing visit for every
  available node type and instead let the Prism::Visitor visit ALL nodes.


glebm (approved):
Nice!

PR #629, opened 2025-04-21
Add check for reserved interpolation keys

jbockler:
We recently ran into `I18n::ReservedInterpolationKey` errors because we used `format` as an interpolation key. That made us wonder why `i18n-tasks`, which we run in CI, didn’t catch it.

Turns out this kind of check isn’t part of the gem – so I added one.

Would love your thoughts on whether this makes sense as a general feature or if we should just keep it as a project-specific check.


glebm (approved):

davidwessman (commented):
Please fix rubocop errors and normalize the new translations 🙂 
You can ignore the ruby 2.6 tests

PR #630, opened 2025-05-04
OpenAI: Fixes using language names

### LINKED ISSUE #617: OpenAI treats uk locale as English instead of Ukrainian
Currently, the prompt reads:

> You are a professional translator that translates content from the %{from} locale to the %{to} locale in an i18n locale array.

This could be improved by expanding the locale key to its full name—for example, en → English, uk → Ukrainian, etc. I've implemented this, and it works.

Another option would be to add context that the key follows the ISO_3166-1_alpha-2 standard (though I haven’t tested this yet).

davidwessman:
It fails to understand that `uk` is Ukrainian and not Great Britain.
Also fixes google translate test, lets not use `!` in the cases where it is not 100% if it will add the correct Spanish use.
Fixes #617


davidwessman:
@dankimio Can you test it out if this works for your usecase?

lib/i18n/tasks/data/language_names.rb
```diff
@@ -0,0 +1,202 @@
+# frozen_string_literal: true
+
+module I18n
+  module Tasks
+    module Data
+      module LanguageNames # rubocop:disable Metrics/ModuleLength
+        # Data loaded from https://github.com/grosser/i18n_data
+        # I18nData.languages.transform_keys!(&:to_sym).transform_values!{|v| v.split(";").first }
```
davidwessman, line 8: @glebm or should we add `i18n_data` as a optional dependency which is only used if it is available?

glebm: I think it's better to have it here as it might need to be expanded it in the future to handle region codes, such as `zh-TW` as "Taiwanese Mandarin"`.

glebm (approved):

PR #632, opened 2025-05-04
Prism: Uses built-in comments handling

davidwessman:
- Avoids the logic for connecting commens to the correct line,
  leave it to Prism.
- Changes the behavior for commens added without e.g. a method call
  on a following line. It will then be connected to the previous line.

### DIFF
--- a/spec/prism_scanner_spec.rb
+++ b/spec/prism_scanner_spec.rb
@@ -346,9 +346,9 @@ RSpec.describe 'PrismScanner' do
         expect(occurrence.path).to eq(
           'spec/fixtures/used_keys/app/controllers/a.rb'
         )
-        expect(occurrence.line_num).to eq(7)
+        expect(occurrence.line_num).to eq(4)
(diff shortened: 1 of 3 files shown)


spec/prism_scanner_spec.rb
```diff
@@ -346,9 +346,9 @@ def value
         expect(occurrence.path).to eq(
           'spec/fixtures/used_keys/app/controllers/a.rb'
         )
-        expect(occurrence.line_num).to eq(7)
+        expect(occurrence.line_num).to eq(4)
         expect(occurrence.line).to eq(
-          "# i18n-tasks-use t('translation.from.comment3')"
+          "I18n.t(\"scoped.translation.\#{variable}\")"
```
glebm, line 351: This makes it so it's not longer as obvious in the report that the calls are coming from a magic comment.
Not sure how big of a deal that is.

glebm (approved):

PR #633, opened 2025-05-05
Specs: Fixes google_translate_spec

davidwessman:
- Removes `!` from the test data where the result may vary depending
  on linebreaks.


davidwessman:
I fixed the linting errors in #634

glebm (approved):

PR #641, opened 2025-05-17
Use configured glossary ID as fallback for Deepl Free plan

moguls753:
Problem: When using the DeepL API Free plan, `DeepL.glossaries.list` always returns `nil`—even if you’ve defined one or more IDs in your i18n-config.yml—because glossary management isn’t supported on Free. As a result, i18n-tasks can’t apply any glossaries.

Solution: Fall back to the IDs in i18n-config.yml when the API returns no glossaries.

### DIFF
--- a/lib/i18n/tasks/translators/deepl_translator.rb
+++ b/lib/i18n/tasks/translators/deepl_translator.rb
@@ -111,9 +111,12 @@ module I18n::Tasks::Translators
       end
     end
 
+    # The Free API endpoint doesn’t expose glossaries via DeepL.glossaries.list,
+    # so if no API-backed glossary is found, fall back to the first ID from i18n-config.yml.
     def options_with_glossary(options, from, to)
-      glossary = find_glossary(from, to)
-      glossary ? { glossary_id: glossary.id }.merge(options) : options
+      configured = @i18n_tasks.translation_config[:deepl_glossary_ids]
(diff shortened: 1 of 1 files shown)


glebm (approved):

lib/i18n/tasks/translators/deepl_translator.rb
```diff
@@ -112,8 +112,9 @@ def configure_api_key!
     end
 
     def options_with_glossary(options, from, to)
-      glossary = find_glossary(from, to)
-      glossary ? { glossary_id: glossary.id }.merge(options) : options
+      configured = @i18n_tasks.translation_config[:deepl_glossary_ids]
```
davidwessman, line 115: It would be nice with some documentation for this edge case. Would @glebm prefer it as a code comment or should we add a section to the Readme?

glebm: Either one is fine but I agree that it'd be good to have this documented

davidwessman: @moguls753 Will you please add a comment describing the case?

moguls753: That's a good idea! I added a comment above.

davidwessman (approved):

PR #728, opened 2026-05-01
Fix crash when human_attribute_name is called with interpolated string argument

jaredmoody:
## Problem

When using `prism: "rails"` scanner config, calling `Model.human_attribute_name` with an interpolated string argument crashes with a `TypeError`:

```
TypeError: no implicit conversion of Array into String
```

**Root cause:** `ArgumentsVisitor` has no handler for `InterpolatedStringNode`. Prism::Visitor's default `visit_child_nodes` returns the raw child node array, which propagates as the `attribute_name` value. In `rails_handle_human_attribute_name`, the array is then used in string concatenation, causing the crash.

The same issue exists in the whitequark AST path: `RailsModelMatcher` did not guard against non-static attribute arguments, producing bogus keys like `activerecord.attributes.model.#<Prism::StringNode:...>` in the missing keys report.

## Fix

1. **`ArgumentsVisitor`**: add explicit `nil`-returning handlers for `InterpolatedStringNode`, `InterpolatedSymbolNode`, and `InterpolatedXStringNode`. The `nil` is compacted away by `process_arguments`, causing the call to be correctly skipped as unresolvable.

2. **`RailsModelMatcher`**: guard against non-`str`/`sym` attribute arguments before attempting key construction.

## Test

Added `Product.human_attribute_name("status.#{status}")` to the shared fixture and a dedicated spec asserting no crash and no spurious key produced.

### DIFF
--- a/lib/i18n/tasks/scanners/ast_matchers/rails_model_matcher.rb
+++ b/lib/i18n/tasks/scanners/ast_matchers/rails_model_matcher.rb
@@ -20,6 +20,10 @@ module I18n::Tasks::Scanners::AstMatchers
 
       value = children[2]
 
+      # Only process static string/symbol arguments; dynamic args (dstr/dsym) cannot be
+      # statically resolved and would produce a bogus key (e.g. "activerecord.attributes.product.").
+      return unless value && %i[str sym].include?(value.type)
+
       model_name = underscore(receiver.to_a.last)
       attribute = extract_string(value)
       # Convert dots to slashes to match Rails' human_attribute_name behavior
--- a/lib/i18n/tasks/scanners/prism_scanners/arguments_visitor.rb
+++ b/lib/i18n/tasks/scanners/prism_scanners/arguments_visitor.rb
@@ -39,6 +39,20 @@ module I18n::Tasks::Scanners::PrismScanners
       node.content
     end
 
+    # Interpolated nodes (e.g. `Product.human_attribute_name("status.#{status}")`) contain dynamic content that cannot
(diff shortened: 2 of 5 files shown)


lib/i18n/tasks/scanners/prism_scanners/arguments_visitor.rb
```diff
@@ -39,6 +39,18 @@ def visit_string_node(node)
       node.content
     end
 
+    def visit_interpolated_string_node(_node)
```
davidwessman, line 44: Could we add some comment here why we return nil? With an example would be great 😊

lib/i18n/tasks/scanners/ast_matchers/rails_model_matcher.rb
```diff
@@ -20,6 +20,8 @@ def human_attribute_name_to_key_occurences(send_node:, location:)
 
       value = children[2]
 
+      return unless value && %i[str sym].include?(value.type)
```
davidwessman, line 25: Does it crash in this matcher too?

davidwessman:
Thank you for making a pr 😊

lib/i18n/tasks/scanners/ast_matchers/rails_model_matcher.rb
jaredmoody: It doesn't crash here, but the same interpolated case isn't handled correctly. I will separate into its own commit and add a comment for clarity.

davidwessman (approved):

Issue #738, opened 2026-05-13
Prism does not properly handle ActiveModels


JochenLutz:
We have some models in our codebase that are not `ActiveRecord`, but `ActiveModel`. Therefore, their attribute localizations are in `active_model.attributes.<model_class>`.

The prism scanner does not handle that, but reports missing keys in `active_record.attributes.<model_class>`.

JochenLutz:
I'm not sure if I should classify that as another case of this issue, or a separate one:

We use [AttrJson](https://github.com/jrochkind/attr_json) and its {{AttrJson::Model}}  to map Postgres's JSON columns to Rails models. {{AttrJson::Model}} builds on {{ActiveModel}}.

But the name resolution is different. Given the following model

```
module AttrJsonModels
  module Foo
    class Bar
      include AttrJson::Model

      attr_json :attribute, :integer
    end
  end
end
```

Then `AttrJsonModels::Foo::Bar.human_attribute_name(:attribute)` resolves to `activemodel.attributes.attr_json_models/foo/bar.attribute`. But the prism scanner falls back to `active_record.attributes.attribute`.

davidwessman:
Yeah, `AttrJson:Model` might be out of scope and would require some different plugin of some sort (there is no such thing for the Prism parser yet).

Could you write an example of the first failing ActiveModel-class as well?

Wonder if we might need something like https://github.com/shopify/rubydex to handle all of these cases 🤔 

Very open to PRs 🙂

JochenLutz:
For now just a PR with a test reproducing the issue. 🙂

JochenLutz:
As you suggest a separate plugin for `AttrJson::Model`, what do you think on adding a plugin API to I18nTasks?

My [I18nTasks::Plugin::ViewComponent](https://github.com/ionos-cloud/i18n-tasks-view-component) currently just hooks into the internals of I18nTasks. As it does not really go deep into the internals, it works (although it is not really nice and might break on updates). But for a plugin handling `AttrJson::Model` I fear I have to dig deep into the internals. And for something that is work in progress like the prism scanner, this feels wrong.

I imagine an API where plugin could register for well-defined hooks. Something akin to [ActiveSupport::Callbacks](https://devdocs.io/rails~8.1/activesupport/callbacks). So the plugin could register for the hook handling the mapping from detected class to key path and modify the result if appropriate.

PR #744, opened 2026-06-16
[Prism Scanner] fix wrongfully reported missing keys in nested models

JochenLutz:
Prism scanner only respects the last part of nested model names. But for [namespaced models](https://guides.rubyonrails.org/active_record_basics.html#creating-namespaced-models) Rails generates keys with the full model name, joined by slashes (`foo/bar` for a `class Foo::Bar`).

When investigating [my issue with ActiveModels that use AttrJsonModel](https://github.com/glebm/i18n-tasks/issues/738) I found this to be an issue also affecting `ActiveRecord` based models when they are namespaced.


davidwessman (approved):
Looks good, please fix rubocop errors

Issue #753, opened 2026-08-10
Implement caching


rwojsznis:
Hey there; first of all - I _love_ i18n-tasks, there was even a dark time when I tried to implement my own i18n-management tool and then I discovered this gem <3

Unfortunately performance on even small/medium-sized (3 locales, 80k keys, 500 files is not great - `i18n-tasks health` takes a around ~35s on M1 Max (using `prism`). I tried to look under the hood and there are not a lot of low hanging fruits. I even went far into checking parallelization using [ractors](https://docs.ruby-lang.org/en/3.4/ractor_md.html), lol.

There is one thing worth considering - and this is what `rubocop` implemented as it suffered from performance issues as well - simple file-based caching. (Of course that is a matter of expiring old entries, as with cache _now you have two problems_)

I AI-sloped some PoC locally after going into multiple dead ends and it seems it could work, but obviously I don't want to shove some LLM-generated PR into your throats xd.

So my question is simply - wdyt? Would you (you - current repo owners) consider adding caching? Do you accept new PRs witch such (quite big and potentially risky) changes? 🤔

glebm:
We did have some parallelism in the past but it was causing issues and was removed in https://github.com/glebm/i18n-tasks/pull/687
I wonder if there is a way to bring it back. This may be tricky because we don't have regression tests for the issues that were caused by parallelism.

File reads are already cached via https://github.com/glebm/i18n-tasks/blob/main/lib/i18n/tasks/scanners/files/caching_file_reader.rb (possibly pointless because the OS also caches file reads).

Given that every device has multiple cores these days, I think parallelism is probably preferable to caching.

rwojsznis:
Thanks for quick reply! I wasn't aware there were some attempts in the past around parallelism, good to know. Personally I didn't had much luck tho - seems like majority of work is cup-bound so GIL quickly suppresses any gains (thus I mentioned the Ractors experiments route). I will poke around some more, there is a chance I missed something 🤔

rwojsznis:
Ok so indeed concurrency (processes) would be the right solution (_thanks GIL_); but I ran some benchmarks "on my app" and in my use case - caching would be actually way more useful - I run `i18n-tasks` _a lot_ - it's a pre-push hook, but for majority of the time it doesn't really _need to_ do anything - so caching would be better - for me 😄 

I think we could shave probably shave like x <times of number of locales you're using in the app> but going beyond that would be very difficult.

[...]

PR #754, opened 2026-08-17
Per process locales normalization

rwojsznis:
Relates to discussion in https://github.com/glebm/i18n-tasks/issues/753#issuecomment-5300995949

---

Quick PoC of running normalization check in processes (if supported); on small app with 3 locales:

### Before

`bundle exec i18n-tasks check-normalized # 25.7s`

### After

`bundle exec i18n-tasks check-normalized # 9.4s`
(which kinda makes sense, should be around `<number of locales> times` faster)

I ran into various dead ends with threads (GIL) / Ractors (a lot of quirks with some shared class variables issue on... activesupport level I think? plus whole API seems to change quite often) and this is AI slopped PoC which gives you the general _idea_; I checked out from `1.1.2` tags as `main` was giving me some prism-related errors.

@glebm wdyt? should we explore this further? 🤔


glebm:
I haven't used Ruby myself in many years, so I'll defer to @davidwessman

davidwessman:
I am on parental leave and have not been at the computer for a while.

Will try to look at it when I am back at the desk in 2 weeks.

rwojsznis:
[closing](https://github.com/glebm/i18n-tasks/issues/753#issuecomment-5421559535) - not gonna pursue this approach after all, sorry for the noise (intentions were honest tho!)

good luck (?) on your parental leave David and congrats! 😄

davidwessman:
Thank you for trying it out 😊

